PR #33, opened 2026-03-29
Add fullscreen screenshot mode for safe area and edge verification

jfversluis:
## Problem

When an AI agent uses DevFlow to verify layout near screen edges, it takes a default screenshot that captures only the **Page content area** — excluding the status bar, home indicator, notch/Dynamic Island, and safe area regions. The agent then concludes the layout "looks fine" because it literally cannot see the device chrome the content may be overlapping with.

The in-app HTTP API already has a `fullscreen=true` query parameter on `/api/screenshot` that triggers composited full-device capture (all UIWindows on iOS, CGWindowListCreateImage on macOS), but this parameter was **never wired through** to:
- `AgentClient.ScreenshotAsync()`
- The `maui_screenshot` MCP tool
- The CLI `MAUI screenshot` command

The CLI had a separate workaround for iOS (`simctl io screenshot` fallback), but the MCP tool path — which is what AI agents use — had no way to request a full-device capture.

### Code (parameter threading)
| Layer | Change |
|-------|--------|
| **AgentClient** | New `bool fullscreen = false` parameter on `ScreenshotAsync()`, sends `fullscreen=true` query param |
| **MCP Tool** | New `fullscreen` parameter with rich description guiding agents to use it for edge/safe area checks |
| **CLI** | New `--fullscreen` option on `MAUI screenshot` command |
| **Output** | JSON output includes `fullscreen` field; text output shows `(fullscreen)` label; MCP returns mode label |

`fullscreen` defaults to `false` to preserve token efficiency — full-device screenshots are larger and most verification doesn't need them.

### Skill file (AI agent guidance)
- **New "Safe Area & Edge Verification" section** — explains the default vs fullscreen difference and lists trigger keywords for when agents should automatically use fullscreen:
  - Safe areas, insets, `UseSafeArea`
  - Content clipping/cutoff at screen edges
  - Status bar overlap, notch clearance, Dynamic Island
  - Home indicator, bottom bar, navigation bar overlap
  - Rounded corner clipping
  - Toolbar/tab bar positioning relative to device chrome
  - Keyboard avoiding behavior
- **Updated command reference table** — documents both modes
- **Token usage tips** — mentions `--fullscreen` alongside element-level screenshots
- **Inspection flow** — added fullscreen example

## Testing

The `fullscreen=true` HTTP API parameter already worked end-to-end (it's been in `DevFlowAgentService.HandleScreenshot` since inception). This PR only wires it through the public API layers. CI will validate compilation.

### DIFF
--- a/.claude/skills/maui-ai-debugging/SKILL.md
+++ b/.claude/skills/maui-ai-debugging/SKILL.md
@@ -200,6 +200,7 @@ if the build fails.
 7. `maui-devflow MAUI screenshot --output screen.png` — visual verification (auto-scaled to 1x on HiDPI)
 8. `maui-devflow MAUI screenshot --id <elementId> --output el.png` — element-only screenshot
 9. `maui-devflow MAUI screenshot --selector "Button" --output btn.png` — screenshot by CSS selector
+10. `maui-devflow MAUI screenshot --fullscreen --output full.png` — full device display including status bar and safe areas
 
 **Property inspection** is more reliable than screenshots for verifying exact runtime values:
 ```bash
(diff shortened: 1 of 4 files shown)


jfversluis:
@Redth not sure what the right approach is here. My scenario was: I was building an app on iOS and the statusbar/safe area was not the color I wanted. So I told Copilot: hey I want the color underneath to also continue upwards in the safe area. And it would take a screenshot and be like: the color is already all the way in the top, so nothing to do here!

Probably because it took a partial screenshot, which is good for efficiency, but not if you're looking at edges or safe areas or things :D

So did an attempt to "fix" that

jfversluis:
Lets skip for now and potentially revisit later

PR #73, opened 2026-04-10
Fix Android install bootstrap when sdkmanager is missing

### LINKED ISSUE #72: Fix circular E2102 failure in maui android install
## Summary
`maui android install` can fail with a circular `E2102` error when an Android SDK directory exists but the command-line tools were never fully installed.

## Repro
1. Have a partial SDK folder at `%LOCALAPPDATA%\\Android\\sdk` with no working `sdkmanager`.
2. Run `maui android install`.
3. Observe the self-referential error:
   ```text
   Error [E2102]: SDK Manager not found. Run ''maui android install'' first.
   ```

## Expected
`maui android install` should bootstrap the Android command-line tools and continue setup instead of reporting a loop.

## Root cause
The CLI considered the SDK installed based on directory presence, then tried to query packages before a real `sdkmanager` executable was available.

rmarinho:
## Summary
Fixes the circular `E2102` failure from `maui android install` when the Android SDK directory exists but the command-line tools were not actually bootstrapped.

## Changes
- resolve `sdkmanager` from real `cmdline-tools` layouts
- treat partial SDK folders as incomplete
- skip early interactive package discovery until the SDK tools are available
- add regression coverage for the incomplete SDK case

Fixes #72

### DIFF
--- a/src/Cli/Microsoft.Maui.Cli.UnitTests/AndroidProviderTests.cs
+++ b/src/Cli/Microsoft.Maui.Cli.UnitTests/AndroidProviderTests.cs
@@ -27,10 +27,28 @@ public class AndroidProviderIsSdkInstalledTests : IDisposable
 	/// <summary>
 	/// Mirrors the <c>AndroidProvider.IsSdkInstalled</c> check so tests stay in sync with the implementation.
 	/// </summary>
-	static bool IsSdkInstalledCheck(string? sdkPath) =>
-		!string.IsNullOrEmpty(sdkPath)
-		&& Directory.Exists(sdkPath)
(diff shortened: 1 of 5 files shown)


rmarinho:
Addressed the review feedback in c5c2962: the health check now keeps \Android SDK\ and \SDK Manager\ as separate concerns so the partial-install case still reports E2102, and sdkmanager resolution now captures the SDK path once before resolving.

rmarinho:
@copilot  is this something that could be better handle on dotnet/android-tools that we consume ?

rmarinho:
@copilot can please  you create a issue on dotnet/android-tools for this fixed there too.

jfversluis (approved):

PR #91, opened 2026-04-16
Port Maui.Gtk PR #19: refactor CairoPlatformImage stream handling

Redth:
Ports the Cairo image stream and pixbuf refactor from the old Maui.Gtk repo.

Original PR: https://github.com/Redth/Maui.Gtk/pull/19

This removes the temp-file-heavy image stream path in favor of Pixbuf-backed load/save helpers and the matching GTK dependency wiring.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -50,6 +50,7 @@
     <PackageVersion Include="Platform.Maui.Linux.Gtk4" Version="$(PlatformMauiLinuxGtk4Version)" />
     <PackageVersion Include="Platform.Maui.Linux.Gtk4.BlazorWebView" Version="$(PlatformMauiLinuxGtk4BlazorWebViewVersion)" />
     <PackageVersion Include="Platform.Maui.Linux.Gtk4.Essentials" Version="$(PlatformMauiLinuxGtk4EssentialsVersion)" />
+    <PackageVersion Include="GirCore.GdkPixbuf-2.0" Version="0.7.0" />
   </ItemGroup>
 
   <!-- Testing (Microsoft.NET.Test.Sdk and xunit.runner.visualstudio are
--- a/platforms/Linux.Gtk4/src/Linux.Gtk4/Graphics/CairoPlatformImage.cs
+++ b/platforms/Linux.Gtk4/src/Linux.Gtk4/Graphics/CairoPlatformImage.cs
@@ -1,4 +1,3 @@
-using System.Runtime.InteropServices;
 using Microsoft.Maui.Graphics;
 
 namespace Microsoft.Maui.Platforms.Linux.Gtk4.Graphics;
@@ -16,8 +15,8 @@ internal class CairoPlatformImage : global::Microsoft.Maui.Graphics.IImage
 
 	internal Cairo.ImageSurface Surface { get; }
 
-	public float Width => cairo_image_surface_get_width(Surface.Handle.DangerousGetHandle());
-	public float Height => cairo_image_surface_get_height(Surface.Handle.DangerousGetHandle());
+	public float Width => Surface.Width;
+	public float Height => Surface.Height;
 
 	public global::Microsoft.Maui.Graphics.IImage Downsize(float maxWidthOrHeight, bool disposeOriginal = false)
 	{
@@ -101,17 +100,8 @@ internal class CairoPlatformImage : global::Microsoft.Maui.Graphics.IImage
 
 	public void Save(Stream stream, ImageFormat format = ImageFormat.Png, float quality = 1)
 	{
-		var tmpPath = System.IO.Path.Combine(System.IO.Path.GetTempPath(), $"maui_img_{Guid.NewGuid():N}.png");
-		try
-		{
-			cairo_surface_write_to_png(Surface.Handle.DangerousGetHandle(), tmpPath);
-			using var fs = File.OpenRead(tmpPath);
-			fs.CopyTo(stream);
-		}
-		finally
-		{
-			try { File.Delete(tmpPath); } catch { }
-		}
+		using var pixbuf = Surface.CreatePixbuf();
+		pixbuf.SaveToStream(stream, format);
 	}
 
 	public Task SaveAsync(Stream stream, ImageFormat format = ImageFormat.Png, float quality = 1)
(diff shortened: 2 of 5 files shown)


platforms/Linux.Gtk4/src/Linux.Gtk4/Graphics/PixbufExtensions.cs
```diff
@@ -0,0 +1,168 @@
...
+		{
+			var bytesRead = stream.Read(buffer, 0, bufferSize);
+			if (bytesRead == 0)
+				break;
+
+			loader.Write(buffer.AsSpan(0, bytesRead));
+		}
+
+		loader.Close();
+		return loader.GetPixbuf();
```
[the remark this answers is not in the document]
Redth: Good catch. I updated `LoadFromStream` to return a copied `Pixbuf` so the result keeps an independent lifetime after the loader is disposed.

platforms/Linux.Gtk4/src/Linux.Gtk4/Graphics/PixbufExtensions.cs
```diff
@@ -0,0 +1,168 @@
...
+
+	public static Pixbuf? CreatePixbuf(this ImageSurface? surface)
+	{
+		if (surface is null)
+			return null;
+
+		var surfaceData = surface.GetData();
+		var bytesPerPixel = surface.Format == Format.Argb32 ? 4 : 3;
+		var pixbufData = new byte[surfaceData.Length / 4 * bytesPerPixel];
+
```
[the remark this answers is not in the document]
Redth: Good catch. I updated CreatePixbuf to flush the Cairo surface before reading its pixel buffer so Save() and other callers consistently serialize the latest contents.

platforms/Linux.Gtk4/src/Linux.Gtk4/Graphics/CairoPlatformImage.cs
```diff
@@ -101,17 +100,8 @@ public void Draw(global::Microsoft.Maui.Graphics.ICanvas canvas, RectF dirtyRect
 
 	public void Save(Stream stream, ImageFormat format = ImageFormat.Png, float quality = 1)
 	{
```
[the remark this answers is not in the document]
Redth: Good catch. The flush is already handled centrally in `CreatePixbuf()`, which is the path `Save()` uses before reading the surface memory, so no additional call is needed here.

Directory.Packages.props
```diff
@@ -50,6 +50,7 @@
     <PackageVersion Include="Platform.Maui.Linux.Gtk4" Version="$(PlatformMauiLinuxGtk4Version)" />
     <PackageVersion Include="Platform.Maui.Linux.Gtk4.BlazorWebView" Version="$(PlatformMauiLinuxGtk4BlazorWebViewVersion)" />
     <PackageVersion Include="Platform.Maui.Linux.Gtk4.Essentials" Version="$(PlatformMauiLinuxGtk4EssentialsVersion)" />
+    <PackageVersion Include="GirCore.GdkPixbuf-2.0" Version="0.7.0" />
```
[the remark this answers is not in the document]
Redth: It's being used in the Linux.Gtk4.csproj file.

platforms/Linux.Gtk4/src/Linux.Gtk4/Graphics/CairoPlatformImage.cs
```diff
@@ -121,38 +111,19 @@ public Task SaveAsync(Stream stream, ImageFormat format = ImageFormat.Png, float
...
-		{
-			try { File.Delete(tmpPath); } catch { }
-		}
+		using var pixbuf = PixbufExtensions.LoadFromStream(stream);
+		if (pixbuf is null)
+			return null;
+
+		var surface = new Cairo.ImageSurface(Cairo.Format.Argb32, pixbuf.Width, pixbuf.Height);
+		using var cr = new Cairo.Context(surface);
+		cr.PaintPixbuf(pixbuf);
```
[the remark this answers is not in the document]
Redth: Good catch. I kept the parameter for API consistency, but it now validates GTK-supported formats and clarifies that GdkPixbuf still auto-detects the actual decoder from the stream contents.

platforms/Linux.Gtk4/src/Linux.Gtk4/Graphics/PixbufExtensions.cs
```diff
@@ -0,0 +1,175 @@
...
+		{
+			pattern.Filter = width > pixbuf.Width || height > pixbuf.Height
+				? Filter.Fast
+				: Filter.Good;
+		}
+
+		context.Paint();
+
+		if (translate || scale)
+			context.Restore();
```
jonathanpeppers: Does this need to be in a `try-finally` block?

platforms/Linux.Gtk4/src/Linux.Gtk4/Graphics/PixbufExtensions.cs
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	public static Pixbuf? LoadFromStream(Stream stream, ImageFormat imageFormat = ImageFormat.Png)
+	{
+		ArgumentNullException.ThrowIfNull(stream);
+		_ = GetImageExtension(imageFormat);
+
+		using var loader = PixbufLoader.New();
+		const int bufferSize = 8192;
+		var buffer = new byte[bufferSize];
```
jonathanpeppers: Can this use `ArrayPool.Rent/Return`? just keep in mind `Rent()` can give you an array larger than requested, so check length appropriately.

Redth: Good call. Switched to ArrayPool<byte>.Shared.Rent/Return with a try/finally, and kept reads/writes bounded to the requested bufferSize (not buffer.Length) since Rent may return a larger array.

platforms/Linux.Gtk4/src/Linux.Gtk4/Graphics/PixbufExtensions.cs
Redth: Good catch - wrapped the Save/Restore pair in a try/finally so the Cairo context state stack stays balanced if Paint or any intermediate call throws.

platforms/Linux.Gtk4/src/Linux.Gtk4/Graphics/PixbufExtensions.cs
```diff
@@ -0,0 +1,189 @@
...
+
+		using var loader = PixbufLoader.New();
+		const int bufferSize = 8192;
+		var buffer = ArrayPool<byte>.Shared.Rent(bufferSize);
+
+		try
+		{
+			while (true)
+			{
+				var bytesRead = stream.Read(buffer, 0, bufferSize);
```
jonathanpeppers: I would use `buffer.Length` here, so if you have a larger buffer (lucky!) you can use it all.

platforms/Linux.Gtk4/src/Linux.Gtk4/Graphics/PixbufExtensions.cs
```diff
@@ -0,0 +1,189 @@
...
+		stream.Write(bytes.GetRegionSpan<byte>(0, bytes.GetSize()));
+	}
+
+	public static Pixbuf? LoadFromStream(Stream stream, ImageFormat imageFormat = ImageFormat.Png)
+	{
+		ArgumentNullException.ThrowIfNull(stream);
+		_ = GetImageExtension(imageFormat);
+
+		using var loader = PixbufLoader.New();
+		const int bufferSize = 8192;
```
jonathanpeppers: So, then you might not even need `bufferSize`, just pass 8192 to `Rent()`.

platforms/Linux.Gtk4/src/Linux.Gtk4/Graphics/PixbufExtensions.cs
Redth: Good call, switched to `buffer.Length` so we use the full rented capacity when the pool hands us a larger array.

platforms/Linux.Gtk4/src/Linux.Gtk4/Graphics/PixbufExtensions.cs
Redth: Agreed, dropped the `bufferSize` const and pass `8192` straight to `Rent()`.

jonathanpeppers (approved):

dalexsoto (approved):

PR #105, opened 2026-04-20
Default Android JDK to 21

rmarinho:
Bumps the default JDK installed/used by the `maui android` CLI from OpenJDK 17 to OpenJDK 21.

### Changes
- `JdkManager.DefaultJdkVersion` 17 → 21
- `maui android install --jdk-version` default → 21
- `maui android jdk install --version` default → 21
- `IJdkManager.InstallAsync` / `IAndroidProvider.InstallJdkAsync` / `InstallAsync` default parameter → 21
- `GetAvailableVersions()` now lists 21 first, then 17
- Health-check fix description updated to reference the default version constant
- Test fakes and stubs updated to match the new signatures

### Compat
- Users who need 17 can still pass `--jdk-version 17` / `--version 17`.
- Existing JDK installs are unaffected; detection still supports OpenJDK 11+.

### Validation
- `dotnet build src/Cli/Cli.slnf` — succeeds
- `dotnet test src/Cli/Microsoft.Maui.Cli.UnitTests` — 268/268 passing

### DIFF
--- a/src/Cli/Microsoft.Maui.Cli.UnitTests/AndroidCommandsTests.cs
+++ b/src/Cli/Microsoft.Maui.Cli.UnitTests/AndroidCommandsTests.cs
@@ -65,6 +65,39 @@ public class AndroidCommandsTests
 	}
 
 	[Fact]
+	public void InstallCommand_JdkVersionDefaultsToDefaultJdkVersion()
+	{
+		// Arrange
+		var androidCommand = AndroidCommands.Create();
+		var installCommand = androidCommand.Subcommands.First(c => c.Name == "install");
+		var jdkVersionOption = (Option<int>)installCommand.Options.First(o => o.Name == "--jdk-version");
+
+		// Act — no --jdk-version supplied, so the default value factory should resolve.
+		var parseResult = installCommand.Parse("install");
+
+		// Assert
+		Assert.Empty(parseResult.Errors);
+		Assert.Equal(Microsoft.Maui.Cli.Providers.Android.JdkManager.DefaultJdkVersion, parseResult.GetValue(jdkVersionOption));
+	}
+
+	[Fact]
+	public void JdkInstallCommand_VersionDefaultsToDefaultJdkVersion()
+	{
+		// Arrange
+		var androidCommand = AndroidCommands.Create();
+		var jdkCommand = androidCommand.Subcommands.First(c => c.Name == "jdk");
+		var jdkInstallCommand = jdkCommand.Subcommands.First(c => c.Name == "install");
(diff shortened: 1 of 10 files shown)


rmarinho:
Rebased onto latest main (clean, no conflicts) and addressed the Copilot review feedback plus findings from two independent review passes (Opus 4.6 and GPT-5.4).

## Changes in this update

**Source of truth** — JdkManager.DefaultJdkVersion is now public const int = 21 and SupportedInstallVersions is a public static readonly IReadOnlyList<int> = { 21, 17 }. Everything defaults from these constants.

**Eliminated duplicated `= 21` defaults** — Optional version parameters are now `int? = null` across:
- `IJdkManager.InstallAsync`
- `IAndroidProvider.InstallJdkAsync` / `InstallAsync`
- `AndroidProvider.InstallJdkAsync` / `InstallAsync` (resolve via `?? JdkManager.DefaultJdkVersion`)
- `FakeJdkManager` / `FakeAndroidProvider` / `StubJdkManager`

**CLI defaults use the constant** — `AndroidCommands.Jdk.cs` and `AndroidCommands.Install.cs` now call `DefaultValueFactory = _ => JdkManager.DefaultJdkVersion`.

**Install validation matches supported download list** — `InstallAsync` now rejects versions not in `SupportedInstallVersions` (previously checked against the 11–21 detection range, which included versions with no download URL). Error message lists the actual supported versions.

**`GetAvailableVersions()` is no longer hardcoded** — now returns `SupportedInstallVersions` directly, so adding a future JDK version is a one-line change.

**Detection no longer artificially caps at JDK 21** — removed `MaxJdkVersion` from the fallback detector. A user with JDK 22/23/etc. installed will now be detected and used (`CheckHealthAsync` already had no upper bound, so the two paths were inconsistent).

**Fake test doubles fixed** — `FakeJdkManager.GetAvailableVersions()` and `StubJdkManager.GetAvailableVersions()` now return `[21, 17]` (matching production), not the stale `[17, 21]`.

**New CLI default tests** — `InstallCommand_JdkVersionDefaultsToDefaultJdkVersion` and `JdkInstallCommand_VersionDefaultsToDefaultJdkVersion` parse the commands with no flag and assert the resolved value equals `JdkManager.DefaultJdkVersion`, so a future accidental revert to 17 (or drift between CLI and manager) would fail.

## Validation
- `dotnet build src/Cli/Cli.slnf` — succeeds (0 errors, 0 warnings)
- `dotnet test src/Cli/Microsoft.Maui.Cli.UnitTests` — **274/274 passing**

Redth (approved):

PR #118, opened 2026-04-21
feat: add expert code review workflows with 3-model adversarial consensus

PureWeen:
## Summary

Adds automated expert code review infrastructure to maui-labs, ported from [PureWeen/PolyPilot](https://github.com/PureWeen/PolyPilot).

## How It Works

Two entry points, one shared review engine:

| Workflow | Trigger | When |
|----------|---------|------|
| `review.agent.md` | `/review` slash command | On-demand re-review |
| `review-on-open.agent.md` | PR opened / ready for review | Automatic on new PRs |

Both import `shared/review-shared.md` which orchestrates:

1. **3 parallel sub-agents** (Opus, Sonnet, Codex) review the PR independently
2. **Adversarial consensus** — 3/3 include, 2/3 include, 1/3 gets challenged then included or discarded
3. **Path/line validation** before posting inline comments (prevents review submission failures)
4. **Final verdict** posted as a COMMENT review with severity-ranked findings

## Expert Reviewer Agent

`.github/agents/expert-reviewer.agent.md` is tuned for DevFlow codebases:
- Multi-targeting patterns (Core vs platform-specific projects)
- CDP/WebSocket protocol correctness
- Thread safety in agent/broker communication
- NuGet packaging configuration
- Platform-specific code organization
- References `.github/copilot-instructions.md` and `.github/skills/maui-platform-backend/SKILL.md`

## Security Design

- **COMMENT-only reviews** (`allowed-events: [COMMENT]`) — `REQUEST_CHANGES` reviews from bots can't be auto-dismissed by subsequent runs, creating stale merge blocks. COMMENT reviews communicate severity through 🔴/🟡/🟢 markers without blocking.
- **`cancel-in-progress: false`** — Prevents non-matching comments from killing in-progress agent runs (slash_command compiles to broad `issue_comment` subscriptions).
- **Shared concurrency group** (`review-<PR>`) — `/review` and auto-review share a group so they don't run simultaneously on the same PR.
- **`hide-older-comments: true`** — Previous review comments are collapsed when a new review runs.
- **`roles: [admin, maintainer, write]`** — Deny-by-default; read-only users cannot trigger reviews.

## Files

| File | Purpose |
|------|---------|
| `.github/agents/expert-reviewer.agent.md` | Review agent definition |
| `.github/workflows/review.agent.md` | `/review` slash command entry point |
| `.github/workflows/review-on-open.agent.md` | Auto-review on PR open |
[...]

PR #120, opened 2026-04-21
ci: add stub workflow for review.agent discovery

PureWeen:
## Summary

Adds a minimal stub `review.agent.lock.yml` to main so that GitHub Actions can discover the `workflow_dispatch` trigger. This enables testing the real review workflow from the feature branch (`feat/expert-review-workflow`, PR #118) using:

```bash
gh workflow run review.agent.lock.yml --ref feat/expert-review-workflow -f pr_number=115
```

### Why this is needed

GitHub requires `workflow_dispatch` workflows to exist on the default branch before they can be triggered via API/CLI/UI. The real workflow only exists on `feat/expert-review-workflow` (PR #118).

### What it does

- The stub does nothing — it exits with failure if run directly
- It only serves as a discovery placeholder so `--ref` can dispatch the real workflow
- Once PR #118 merges, this stub is replaced by the real compiled workflow

### Safe to merge

- `permissions: {}` — no permissions granted
- The stub job just prints a message and exits 1
- No secrets, no checkout, no tools

### DIFF
--- a/.github/workflows/review.agent.lock.yml
+++ b/.github/workflows/review.agent.lock.yml
@@ -1,0 +1,29 @@
+# Stub workflow — enables discovery of review.agent.lock.yml so that
+# `gh workflow run review.agent.lock.yml --ref <branch>` works.
+# The real workflow lives on the feature branch; --ref executes THAT version.
+# This stub is safe to leave on main — it does nothing on its own.
+#
+# Once PR #118 (feat/expert-review-workflow) merges, this stub is replaced
+# by the real compiled workflow and can be removed.
+
+name: "Expert Code Review"
+
+on:
+  workflow_dispatch:
+    inputs:
+      pr_number:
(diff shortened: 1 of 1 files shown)


Redth (approved):

PureWeen:
Re: [Copilot suggestion to change `type: number` to `type: string`](https://github.com/dotnet/maui-labs/pull/120#discussion_r3120069355)

Not a valid concern — `type: number` is a supported `workflow_dispatch` input type (alongside `string`, `boolean`, `choice`, and `environment`). It was added to GitHub Actions after the original four types.

The real workflow on `feat/expert-review-workflow` and the gh-aw compiled lock file both use `type: number` for this input, and the compiler accepted it without warnings. The stub should match the real workflow's schema to ensure consistent behavior when dispatching with `--ref`.

PR #142, opened 2026-04-23
Add debug app identity isolation for DevFlow builds

vitek-karas:
## Summary
- add Debug-only app identity isolation to the DevFlow agent MSBuild targets so DevFlow-enabled builds from separate worktrees or agent sessions do not collide on one installed mobile app identity
- keep the behavior automatic for Debug builds, with opt-out and explicit suffix overrides via MSBuild properties or environment variables
- make the Android integration fixture resolve the built package name dynamically instead of assuming a fixed package id

## How the debug identity is created
When identity isolation is enabled, DevFlow rewrites `ApplicationId` to:

`<baseApplicationId>.debug.<suffix>`

The suffix is created like this:
1. use `MauiDevFlowDebugAppIdentitySuffix` (or `MAUI_DEVFLOW_DEBUG_APP_IDENTITY_SUFFIX`) if it was provided
2. otherwise use the full project path (`MSBuildProjectFullPath`)
3. lowercase it and remove every character except `a-z` and `0-9`
4. keep the last 24 characters if it is longer than that
5. prefix the result with `dw`

Example:
- base app id: `com.example.myapp`
- resolved suffix: `dwagent42`
- final debug app id: `com.example.myapp.debug.dwagent42`

## Testing
Added coverage for:
- default Debug rewrite behavior in `MauiDevFlowAgentTargetsTests`
- explicit suffix override
- opt-out behavior
- Release no-op behavior
- Android integration harness resolution of the built package id from `AndroidManifest.xml`

Validation run:
- `dotnet test src/DevFlow/Microsoft.Maui.DevFlow.Tests/ --filter "FullyQualifiedName~MauiDevFlowAgentTargetsTests"`
- `dotnet test src/Cli/Microsoft.Maui.Cli.UnitTests/ --filter "FullyQualifiedName~ProfileCommandTests"`
- `dotnet test src/DevFlow/Microsoft.Maui.DevFlow.Tests/`
- `dotnet test src/DevFlow/Microsoft.Maui.DevFlow.Agent.IntegrationTests/ --filter "FullyQualifiedName=NoSuchTest"` (build-only validation)

## Question
Should we add a true emulator/simulator test for this next?

Short version of what that would take: introduce a dedicated multi-app integration fixture that builds the same sample twice with different suffixes, installs/launches both identities, and verifies DevFlow can address them separately without the installs colliding.


jfversluis:
Nice work @vitek-karas! One suggestion from the review for a potential follow-up: consider emitting an MSBuild `<Message>` when the ApplicationId is rewritten, so developers aren't surprised when `adb` shows an unexpected package name. Something like:

```xml
<Message Importance="normal"
  Text="DevFlow: ApplicationId rewritten to '$(ApplicationId)' for debug identity isolation (opt out with -p:MauiDevFlowEnableDebugAppIdentityIsolation=false)" />
```

Not  the rest looks solid.blocking

jfversluis (approved):
Reviewed with multi-model deep dive (Opus 4.5, GPT-5.2, Sonnet 4). All three reviewers agree: MSBuild timing is correct, suffix algorithm is sound, NuGet target pattern is right, test coverage is comprehensive. No blocking issues found. One minor follow-up suggestion left as a comment (MSBuild info message).

PR #115, opened 2026-04-21
Add 5 critical MCP tools: capabilities, back, key, gesture, batch

rmarinho:
## Summary

Closes the critical gap between DevFlow's AgentClient API and the MCP tool surface. After the v1 spec alignment (2c05e684), several new AgentClient methods had no MCP tools.

## New MCP Tools

| Tool | File | AgentClient Method | Description |
|---|---|---|---|
| `maui_capabilities` | AgentTools.cs | `GetCapabilitiesAsync` | Discover agent-supported features before using other tools |
| `maui_back` | NavigationTools.cs | `BackAsync` | System back navigation |
| `maui_key` | InteractionTools.cs | `KeyAsync` | Keyboard input (enter, backspace, text typing) |
| `maui_gesture` | InteractionTools.cs | `GestureAsync` | Touch gestures (swipe, tap, longpress) with validation |
| `maui_batch` | BatchTools.cs (new) | `BatchAsync` | Atomic multi-action sequences with JSON validation |

## Design Decisions

- **`maui_gesture`** validates types to the supported backend set (`swipe`, `tap`, `longpress`) and requires `direction` for swipes — fails fast instead of silently succeeding
- **`maui_key`** documents actual backend behavior (enter/backspace work on Entry/Editor/SearchBar; other keys may be no-ops)
- **`maui_batch`** accepts a JSON string with strict validation (array check, object check, action field check) with descriptive error messages
- **`maui_capabilities`** placed in AgentTools since it's agent metadata, not UI interaction

## Remaining Gaps (tracked in issues)

- **Profiler tools** (6 methods): #112
- **WebView interaction tools** (4+ methods): #113

## Testing

- ✅ Build passes (`dotnet build src/Cli/Microsoft.Maui.Cli/`)
- ✅ Existing 260 unit tests unaffected (12 pre-existing failures unchanged)
- Manual testing: use `maui devflow mcp` with a running MAUI app, or the equivalent CLI commands

PR #144, opened 2026-04-24
Harden MCP tools: batch exceptions, gesture null params, capabilities message

jfversluis:
Follow-up to #115.

Three hardening fixes identified during the review of the MCP tools PR:

### 1. Batch exception handling
`agent.BatchAsync()` could throw `HttpRequestException`, `TaskCanceledException`, or `JsonException` when the agent is unreachable or crashes mid-request. Other tools handle this via `PostActionAsync` (which catches internally), but batch uses raw `BatchAsync`. Now wrapped with a try/catch that returns a friendly error.

### 2. Gesture nullable parameter serialization
`GestureAsync` was including null `distance`/`durationMs` in the JSON payload, which would fail on the agent non-nullable DTO (`double Distance = 120`, `int DurationMs = 200`). Now omits null values from the JsonObject so the agent uses its defaults.

### 3. Capabilities error message
Changed from "Agent not responding" (misleading for older agents) to a message that distinguishes connectivity issues from feature support.

## Validation
`dotnet test src/Cli/Microsoft.Maui.Cli.UnitTests/` - 341 passed (5 pre-existing locale-dependent failures in OutputFormatterTests unrelated to this change)

### DIFF
--- a/src/Cli/Microsoft.Maui.Cli/DevFlow/Mcp/Tools/AgentTools.cs
+++ b/src/Cli/Microsoft.Maui.Cli/DevFlow/Mcp/Tools/AgentTools.cs
@@ -96,7 +96,7 @@ public sealed class AgentTools
         var agent = await session.GetAgentClientAsync(agentPort);
         var capabilities = await agent.GetCapabilitiesAsync();
         if (capabilities.ValueKind == System.Text.Json.JsonValueKind.Undefined)
-            return "Agent not responding. Is the app running?";
+            return "Unable to retrieve capabilities. The agent may not support this feature or may be running an older version.";
         return CliJson.SerializeUntyped(capabilities, indented: false);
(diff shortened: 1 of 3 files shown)


jfversluis:
Addressed review feedback in the latest commit:

-  **ScrollAsync nullable  same pattern as GestureAsync, omits null valuesparams** 
-  ** widened catch filter from TaskCanceledExceptionOperationCanceledException** 
-  **Batch error  now mentions both connectivity and version/feature supportmessage** 
-  **Capabilities  now mentions both app-not-running and older versionmessage** 

Resolved the corresponding review threads. The GestureAsync test coverage request is  CI will validate the shape works e2e; happy to add a dedicated unit test in a follow-up.valid

rmarinho (approved):

PR #145, opened 2026-04-24
Add MSBuild info message for DevFlow debug identity isolation

jfversluis:
Follow-up to #142.

Emits a normal-importance MSBuild message when debug identity isolation rewrites the ApplicationId, so developers see the change in build output and know how to opt out.

Example output:
```
DevFlow: ApplicationId rewritten from 'com.example.myapp' to 'com.example.myapp.debug.dw...' for debug identity isolation. Opt out with -p:MauiDevFlowEnableDebugAppIdentityIsolation=false
```

## Changes
- Added `<Message>` task to both `build/` and `buildTransitive/` targets
- Added test verifying the message appears in Debug builds
- Added test verifying the message does NOT appear when isolation is disabled or in Release

## Validation
`dotnet test src/DevFlow/Microsoft.Maui.DevFlow.Tests/ --filter MauiDevFlowAgentTargetsTests`

### DIFF
--- a/src/DevFlow/Microsoft.Maui.DevFlow.Agent/build/Microsoft.Maui.DevFlow.Agent.targets
+++ b/src/DevFlow/Microsoft.Maui.DevFlow.Agent/build/Microsoft.Maui.DevFlow.Agent.targets
@@ -56,5 +56,8 @@
     <ItemGroup>
       <Compile Include="$(_MauiDevFlowPortFile)" />
     </ItemGroup>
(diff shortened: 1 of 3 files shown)


jfversluis:
Addressed review feedback:

-  **Verbosity  `RunSetMauiDevFlowPortTarget` now only defaults to `/v:minimal` when no explicit `/v:` arg is provided, instead of relying on MSBuild's "last switch wins" semanticsfragility** 
-  **Incremental build  acknowledged as acceptable since `Importance="normal"` suppresses it at default verbositymessage** 

Resolved all threads.

Redth:
@jfversluis let's look at #148 as an alternative to the approach that was merged in... Changing the application id has some implications that aren't great, I think this might be a tidier way to accomplish the same goal.

PR #160, opened 2026-04-27
Harden tar extraction validation in JdkManager

rmarinho:
## Summary

Adds pre-extraction validation to `JdkManager.ExtractArchiveAsync()` for the non-Windows (tar) code path, bringing it in line with the safety guarantees already provided by `ZipFile.ExtractToDirectory` on Windows.

## Changes

**`JdkManager.cs`**:
- Added `ValidateArchiveEntries()` — reads the tar.gz archive with `System.Formats.Tar.TarReader` before extraction and validates all entry paths and symlink targets resolve within the destination directory
- Added `StripPathComponents()` helper to simulate `--strip-components=N` during validation
- Uses OS-aware string comparison (case-sensitive on Linux/macOS, case-insensitive on Windows)

**`JdkManagerExtractionTests.cs`** (new):
- 7 unit tests covering safe archives, traversal entries, absolute paths, symlinks (absolute and relative), empty archives, and the StripPathComponents helper

## Risk

Low — validation is read-only and runs before the actual extraction. If the archive cannot be parsed, it falls through to let `tar` report its own error.


src/Cli/Microsoft.Maui.Cli/Providers/Android/JdkManager.cs
```diff
@@ -319,12 +319,17 @@ async Task ExtractArchiveAsync(string archivePath, string targetPath, Cancellati
...
 		{
-			// Use tar for tar.gz extraction
+			// Pre-validate: scan archive entries for path traversal before extracting
+			ValidateArchiveEntries(archivePath, targetPath, stripComponents: 1);
+
+			// tar strips leading '/' from entries by default (unless -P is passed),
+			// preventing absolute-path entries from escaping the destination.
 			var result = await ProcessRunner.RunAsync(
 				"tar",
 				["-xzf", archivePath, "-C", targetPath, "--strip-components=1"],
```
jonathanpeppers, line 335: from AI:
> Both GNU tar and bsdtar (libarchive) strip absolute paths and reject `..` path components by 
default, preventing zip-slip/path-traversal attacks. This has been the default behavior for over a 
decade — you'd have to explicitly opt out with `--absolute-names` or `--insecure` to be vulnerable.
> 
>  tar: Removing leading '/' from member names
>  tar: Member with '..' in name, skipping
> 
>  The command `tar -xzf <archive> -C <target> --strip-components=1` is safe — `-C` constrains 
> extraction to the target directory, and the default security protections remain active.
>  
>  Note: this only applies to shelling out to `tar`. If using archive libraries directly (e.g., .NET 
> `ZipArchive`, Java `TarArchiveInputStream`, or libarchive's C API), you must validate paths 
> yourself — the library APIs don't enable these protections by default.

Should we, in fact:
* Do nothing?
* Provide a way for someone to replace their version of "tar" -- would they just update `$PATH`?
* Error if `tar` is too old

rmarinho:
Huge thanks @jonathanpeppers — you nailed it, and I appreciate you pushing back on this rather than just rubber-stamping it. 🙌

You're absolutely right: shelling out to `tar -xzf -C <target> --strip-components=1` is already safe by default on both GNU tar and bsdtar (leading `/` stripped, `..` members rejected — has been the default for ~20 years). And `ZipFile.ExtractToDirectory` covers the Windows path on .NET 8+. The pre-extraction validation this PR added mostly duplicates those guarantees, and — as the review bots also caught — `System.Formats.Tar.TarReader` normalizes entry names so the absolute-path branch couldn't even observe the attacks it claimed to block.

Going with **Option 1 (do nothing)**. JDK archives come from a trusted HTTPS source (Adoptium/Microsoft), so the marginal hardening isn't worth the extra code path or the maintenance surface.

Closing this out. If we ever start extracting untrusted archives, a cleaner belt-and-suspenders would be to extract into a sibling temp dir and `Directory.Move` on success — happy to revisit then. Thanks again for the careful review!

PR #189, opened 2026-04-29
Merge DevFlow NuGets in ADO pipeline

Redth:
PR #173 taught GitHub Actions to combine the raw macOS and Windows DevFlow package outputs before publishing final Agent and Blazor packages. The Azure DevOps official pipeline still published directly from the Windows DevFlow job, so it needed the same treatment before Arcade validation, BAR publishing, and NuGet.org publishing.

## Approach

- Split the DevFlow official build into macOS and Windows raw package producer jobs.
- Publish those raw outputs as isolated pipeline artifacts instead of final package assets.
- Add a final Windows merge job that downloads both raw package sets, suppresses partial Agent/Blazor packing during the pass-through build, repacks `Microsoft.Maui.DevFlow.Agent` and `Microsoft.Maui.DevFlow.Blazor` from both platform inputs, then runs the normal sign/publish path.
- Leave the Cli and Linux GTK4 official build and NuGet.org publish stages unchanged.

## Notes for reviewers

The new `PackDevFlowPlatformNuGets=false` property only affects the final merge job. Default local and CI packaging behavior for the Agent and Blazor projects remains packable.

## Validation

- Parsed `eng/pipelines/devflow-official.yml` as YAML and verified job names are unique.
- Ran `git diff --check`.
- Verified `PackDevFlowPlatformNuGets=false` changes Agent/Blazor `IsPackable` to `false` while the default remains `true`.
- Verified Arcade package output paths still resolve to `artifacts/packages/Release/Shipping`.

### DIFF
--- a/eng/pipelines/devflow-official.yml
+++ b/eng/pipelines/devflow-official.yml
@@ -73,8 +73,118 @@ extends:
           testResultsFormat: xunit
           enableTelemetry: true
           jobs:
-          - job: DevFlow
-            displayName: DevFlow - Windows
+          # DevFlow Agent/Blazor need macOS-built Apple assets and Windows-built
+          # Windows assets. Keep raw packages isolated, then publish only the
+          # merged packages from the final Windows job.
+          - job: DevFlow_MacRawPackages
+            displayName: DevFlow - macOS raw packages
+            pool:
+              name: Azure Pipelines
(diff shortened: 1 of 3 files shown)


Redth:
Simplifying with another approach.

PR #239, opened 2026-05-06
fix: Add contents:write to Comet CI for release asset uploads

jfversluis:
Release asset uploads fail with 403 because the workflow lacks `contents: write` permission.


jfversluis:
@copilot rebase this PR

PR #286, opened 2026-05-20
Use Assert.Fail in WebView integration test

Redth:
The Windows DevFlow integration suite passes, but the local run surfaced an xUnit analyzer warning in the WebView integration test helper. This change keeps the test behavior the same while using the analyzer-preferred assertion API.

## Summary

- Replaces `Assert.True(false, message)` with `Assert.Fail(message)` in `WebViewTests`.
- Leaves the timeout failure message and fallback flow unchanged.

## Testing

- `DEVFLOW_TEST_PLATFORM=windows dotnet test src\DevFlow\Microsoft.Maui.DevFlow.Agent.IntegrationTests\Microsoft.Maui.DevFlow.Agent.IntegrationTests.csproj` - 113/113 passed
- Re-ran the same Windows integration suite after the fix - 113/113 passed, build succeeded with no warnings

### DIFF
--- a/src/DevFlow/Microsoft.Maui.DevFlow.Agent.IntegrationTests/WebViewTests.cs
+++ b/src/DevFlow/Microsoft.Maui.DevFlow.Agent.IntegrationTests/WebViewTests.cs
@@ -129,7 +129,7 @@ public class WebViewTests : IntegrationTestBase
             return "0";
         }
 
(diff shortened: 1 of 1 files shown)


Redth:
Thanks for the review. No actionable issues were reported, so no changes are needed.

jfversluis (approved):

PR #316, opened 2026-06-08
Add MAUI PR artifact hives

jfversluis:
## Summary
- add `maui project version set --pr/--pull-request` to discover and apply dotnet/maui PR PackageArtifacts
- store downloaded artifacts in managed MAUI hives under `~/.maui/hives/dotnet-maui/pr-<number>/build-<id>/`
- write hive metadata, stage replacements safely, lock per hive, guard zip extraction, and update project NuGet.config/package versions
- add `--hive-path` with `--artifact-path` as a hive-root alias
- add `--target-framework` / `--framework` / `-f` so PR/nightly/stable version changes can update project TFMs safely
- select latest stable/nightly versions compatible with the project/requested target framework when possible
- surface in-progress dotnet/maui PR builds so users know the latest completed artifacts may be stale

## Validation
- `./eng/common/dotnet.sh test src/Cli/Microsoft.Maui.Cli.UnitTests/Microsoft.Maui.Cli.UnitTests.csproj --configuration Debug --no-restore` (`578 passed`)
- packed and installed local `Microsoft.Maui.Cli 0.1.0-dev`
- smoke-tested help/options, explicit `--framework` updates, conditional TFM appends, invalid framework/source failures, inherited/property-composed TFM failures, latest stable compatibility selection, PR dry-runs for `35783`/`35786`, and real PR apply for `35783`
- verified failed PR preflight and failed project-file writes discard staged hives without promoting final artifacts, while successful PR apply promotes the hive and writes metadata

### DIFF
--- a/src/Cli/Microsoft.Maui.Cli.UnitTests/CommandConstructionTests.cs
+++ b/src/Cli/Microsoft.Maui.Cli.UnitTests/CommandConstructionTests.cs
@@ -30,6 +30,10 @@ public class CommandConstructionTests
 		Assert.Contains(versionCommand.Subcommands, command => command.Name == "list");
 		Assert.Contains(versionCommand.Subcommands, command => command.Name == "set");
 		Assert.Contains(versionCommand.Subcommands, command => command.Name == "use-workload");
+		var setCommand = Assert.Single(versionCommand.Subcommands, command => command.Name == "set");
+		Assert.Contains(setCommand.Options, option => option.Name == "--pr");
+		var hivePathOption = Assert.Single(setCommand.Options, option => option.Name == "--hive-path");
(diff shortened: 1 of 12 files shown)


jfversluis:
Addressed the review feedback: the artifact extraction boundary check now uses an OS-appropriate path comparison (case-insensitive on Windows, ordinal elsewhere). This follow-up also adds target-framework compatibility/update support, in-progress PR build warnings, and staged PR hive promotion after preflight validation.

jfversluis:
Addressed the latest review feedback: target framework updates now only touch real MSBuild property elements under PropertyGroup, and PR artifact hives are promoted only after project/NuGet/version writes succeed. Added regression coverage and validated with full CLI tests plus packed-tool smoke for the project-write-failure path.

PR #337, opened 2026-06-18
Stream live progress for `maui android install` package downloads

rmarinho:
## Problem

Running e.g.:

```
maui android install --ci --accept-licenses --packages "...,system-images;android-37.0;google_apis_ps16k;arm64-v8a"
```

made the install **appear to hang**: the Spectre progress bar froze at `Installing system-images;... (6/6) 100%` for many minutes. It was *not* actually hung — the ~1.5GB system-image zip was downloading the whole time (verified: the temp zip under `~/Library/Android/sdk/.temp/` grew steadily and the `sdkmanager` java process stayed in `R` state). It only **looked** hung because no progress was surfaced.

Two root causes:

1. **`SdkManager.InstallPackagesAsync` reported progress too early.** It called `onProgress(pkg, i+1, total)` *before* awaiting the install, so the bar jumped to the package's slot at 100% the instant the install *started*, then sat there until it finished.
2. **The upstream library buffers output.** `Xamarin.Android.Tools.SdkManager` (NuGet `Xamarin.Android.Tools.AndroidSdk`) runs `sdkmanager <pkg>` and captures stdout/stderr into in-memory `StringWriter`s (its private `RunSdkManagerAsync`). It never streams the live `[=====   ] 45% Downloading …` / `… Unzipping …` lines, so our wrapper had nothing to surface. (Confirmed by decompiling the assembly — only the bootstrap/download path exposes `SdkBootstrapProgress`; package installs route through the buffered path.)

## Approach

Add a **streaming** `InstallPackagesAsync` overload in our own `SdkManager` wrapper that keeps the fix entirely in this repo (no upstream release required):

- Resolves the `sdkmanager` path the same way the wrapper already does (`SdkManagerPath`).
- Invokes `sdkmanager <pkg>` directly, one package at a time, with stdout/stderr redirected.
- Streams stdout line-by-line (`ReadLineAsync` treats a lone `\r` as a line terminator, so the in-place progress-bar updates each surface as their own line) and parses each line into `(phase, percent)`.
- Reports a real per-package percentage + phase via a new `AndroidPackageInstallProgress` callback, which drives the existing Spectre live progress bar (Step 4 of `AndroidCommands.Install.cs`). The overall bar now advances as `(completedPackages + currentPackage%) / total`.
- When `acceptLicenses` is `true`, auto-writes `y` to stdin every ~500ms (mirroring the upstream auto-accept loop) so the install stays **non-interactive** with `--accept-licenses`.
- On cancellation, kills the process tree and propagates `OperationCanceledException` (not a wrapped `MauiToolException`); background stdin/stdout/stderr tasks are deterministically stopped and drained before the process is disposed.

JSON / non-interactive mode is untouched (it routes through the existing buffered `InstallAsync`).

## Parser

`SdkManager.TryParseInstallProgressLine(line, out phase, out percent)` is a small pure function, tolerant of:
- progress-bar prefixes `[====   ]`
- carriage-return in-place updates (uses the last segment)
- phase-only lines with no percentage (e.g. `Downloading foo.zip` → percent `-1`)
- percentages clamped to 0–100
- malformed / non-progress lines (returns `false`)

## Public API changes

- **New** `AndroidPackageInstallProgress` record and a new `InstallPackagesAsync(..., Action<AndroidPackageInstallProgress>?, …)` overload on `IAndroidProvider` / `AndroidProvider` / `SdkManager`.
- The existing `Action<string,int,int>` overload is **preserved** (it now delegates to the richer one), so existing callers and tests are unaffected.
- **No shipped NuGet break:** `AgentClient` (the `Microsoft.Maui.DevFlow.Driver` public surface) does not expose package installs. `IAndroidProvider` lives inside the `maui` global tool, not a shipped library.

## Tests

`SdkManagerProgressParserTests` (xUnit, `src/Cli/Microsoft.Maui.Cli.UnitTests/`) covers download %, unzip %, percent-without-phase, phase-without-percent, `\r` in-place updates, clamping, and malformed/no-progress lines. **No network access.**

```
dotnet test src/Cli/Microsoft.Maui.Cli.UnitTests/
Passed!  - Failed: 0, Passed: 589, Skipped: 0, Total: 589
```

A `code-review` pass flagged three process-handling issues (cancellation race wrapping OCE, the stdin loop not being awaited, and abandoned reader tasks on cancel); all three were fixed and re-verified before commit.

## Notes for reviewers

- Draft — please review the streaming/process teardown in `SdkManager.RunSdkManagerInstallAsync` and the overall-bar math in `AndroidCommands.Install.cs`.
- Manual end-to-end verification of a live ~1.5GB system-image download on macOS is still recommended before merge (not exercised by unit tests, by design).

### DIFF
--- a/src/Cli/Microsoft.Maui.Cli.UnitTests/Fakes/FakeAndroidProvider.cs
+++ b/src/Cli/Microsoft.Maui.Cli.UnitTests/Fakes/FakeAndroidProvider.cs
@@ -135,6 +135,18 @@ public class FakeAndroidProvider : IAndroidProvider
 		return Task.CompletedTask;
 	}
 
+	public Task InstallPackagesAsync(IEnumerable<string> packages, bool acceptLicenses, Action<AndroidPackageInstallProgress>? onProgress, CancellationToken cancellationToken = default)
+	{
+		var pkgList = packages.ToList();
+		InstalledPackageSets.Add(pkgList);
+		if (onProgress is not null)
+		{
+			for (var i = 0; i < pkgList.Count; i++)
+				onProgress(new AndroidPackageInstallProgress(pkgList[i], i + 1, pkgList.Count, "Installing", 100));
+		}
+		return Task.CompletedTask;
+	}
+
 	public Task UninstallPackagesAsync(IEnumerable<string> packages, CancellationToken cancellationToken = default)
 	{
 		UninstalledPackageSets.Add(packages.ToList());
--- a/src/Cli/Microsoft.Maui.Cli.UnitTests/SdkManagerProgressParserTests.cs
+++ b/src/Cli/Microsoft.Maui.Cli.UnitTests/SdkManagerProgressParserTests.cs
@@ -1,0 +1,129 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Maui.Cli.Providers.Android;
+using Xunit;
+
+namespace Microsoft.Maui.Cli.UnitTests;
+
+/// <summary>
+/// Tests for <see cref="SdkManager.TryParseInstallProgressLine"/>, the pure parser that turns
+/// streamed <c>sdkmanager</c> stdout lines into (phase, percent) so the install progress bar
+/// reflects real download/extraction progress instead of appearing to hang.
+/// </summary>
+public class SdkManagerProgressParserTests
+{
+	[Fact]
+	public void TryParse_DownloadingLineWithPercent_ReturnsDownloadingAndPercent()
+	{
+		var ok = SdkManager.TryParseInstallProgressLine(
+			"[=========                    ] 33% Downloading system-image.zip",
+			out var phase, out var percent);
+
+		Assert.True(ok);
+		Assert.Equal("Downloading", phase);
+		Assert.Equal(33, percent);
+	}
+
+	[Fact]
+	public void TryParse_UnzippingLineWithPercent_ReturnsUnzippingAndPercent()
+	{
+		var ok = SdkManager.TryParseInstallProgressLine(
+			"[====================         ] 78% Unzipping... google_apis/arm64-v8a/system.img",
+			out var phase, out var percent);
+
+		Assert.True(ok);
+		Assert.Equal("Unzipping", phase);
+		Assert.Equal(78, percent);
+	}
+
+	[Fact]
+	public void TryParse_PercentWithoutSpaceBeforeSign_IsParsed()
+	{
+		var ok = SdkManager.TryParseInstallProgressLine("[=====] 5 % Downloading", out var phase, out var percent);
+
+		Assert.True(ok);
+		Assert.Equal("Downloading", phase);
+		Assert.Equal(5, percent);
+	}
+
+	[Fact]
+	public void TryParse_PercentWithUnknownTrailingText_ReturnsPercentWithEmptyPhase()
+	{
+		var ok = SdkManager.TryParseInstallProgressLine("[==========          ] 50%", out var phase, out var percent);
+
+		Assert.True(ok);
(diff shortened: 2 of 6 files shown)


rmarinho:
Thanks for the review — addressed all 7 findings in `0aa88e61`:

1. 🟡 **Backwards progress bar** — the overall bar is now clamped to a high-water mark, so the Downloading 100% → Unzipping 0% transition can no longer move it backwards.
2. 🟡 **`SafeAwait` swallowing stdout faults** — on a clean (non-cancelled) exit, a faulted stdout reader (e.g. the progress callback throwing) is now rethrown via `ExceptionDispatchInfo` instead of being silently swallowed.
3. 🟡 **`acceptLicenses=false` hang** — the streaming path now falls back to the buffered upstream install whenever licenses can't be auto-accepted, so a license prompt can't deadlock behind the redirected stdout.
4. 🟢 **Legacy adapter multi-fire** — the `Action<string,int,int>` adapter now collapses the per-line stream back to one call per package, preserving its original contract.
5. 🟢 **Regex on hot path** — switched to a `[GeneratedRegex]` partial method.
6. 🟢 **Redundant `SyncPaths()` + dead null-guard** — cache `SdkManagerPath!` after `EnsureAvailable()` and dropped the dead `?? throw`.
7. 🟢 **Misleading `\r` comment** — clarified that the carriage-return split is defensive handling for direct callers; the streaming path already splits on `\r` via `ReadLineAsync`.

Build clean, all 589 CLI unit tests green.

jfversluis (dismissed):

rmarinho:
Addressed the round-2 review feedback in `ffe4b90d`:

| Finding | Resolution |
|---|---|
| 🟡 Progress callback ignored when `acceptLicenses=false` | Fixed. `InstallPackagesAsync` now loops per package even without auto-accept: it emits the seed callback for each package and falls back to the buffered upstream install per package, so callers still get coarse per-package progress instead of a silent single buffered call. |
| 🟡 Pipe-buffer deadlock if `onProgress` throws | Fixed. We now race `WaitForExitAsync` against the stdout reader (`Task.WhenAny`); if the reader faults first we kill the process tree so stdout stops filling the OS pipe buffer. The existing post-drain `ExceptionDispatchInfo` rethrow still surfaces the actual fault exactly once. |
| 🟡 Windows `.bat` under `UseShellExecute=false` | Applied defensively: on Windows, when the resolved sdkmanager ends in `.bat`, we invoke it via `cmd.exe /c <path>`. (Noting the earlier bot "disagree" — this is a low-cost hardening that avoids relying on `.bat` being directly executable as a PE image.) |
| 🟡 Missing `--sdk_root` | Fixed. We now pass `--sdk_root=<sdkPath>` explicitly, mirroring the upstream wrapper, so installs target the resolved SDK even when `ANDROID_HOME`/`ANDROID_SDK_ROOT` aren't set in the ambient env. |
| 🟢 Redundant `SyncPaths()` via `SdkManagerPath` | Fixed. Resolve the sdkmanager path once from the already-synced `sdkPath` (`ResolveSdkManagerPath(sdkPath) ?? _sdkManager.FindSdkManagerPath()!`). |
| 🟢 Parser tests | Fixed. Split the fragile `if (line.Contains('%'))` theory into two unambiguous tests (no-progress lines return false; a bare progress bar returns percent + empty phase), and renamed the whitespace-tolerance test to reflect its `"5 %"` input. |
| Copilot: misleading stop-token comment | Fixed. Comment now says the CTS is linked to the caller's token (it is). |

Build is clean and all 590 CLI unit tests pass (rebased on top of the 1.0.189-preview.58 `Xamarin.Android.Tools.AndroidSdk` bump from main).

No action taken on three findings the reviewer flagged as no-consensus/likely-noise (faulted-reader check "fragility", the regex `%` false-positive concern, and the `FakeAndroidProvider` seed `Percent` value) — happy to revisit if you feel strongly.

jfversluis (approved):

PR #348, opened 2026-06-23
Check GTK version at startup

noiseonwires:
The Linux GTK4 backend uses APIs that only exist in newer GTK releases — gtk_css_provider_load_from_string (GTK 4.12+) and Gtk.FileDialog (GTK 4.10+). On older GTK runtimes these resolve to missing native entry points, so the app crashes with a cryptic EntryPointNotFoundException deep inside the first handler that loads CSS.

This PR adds a fast, fail-early version check in GtkMauiApplication.Run that validates the installed GTK runtime before any handlers run.

### DIFF
--- a/platforms/Linux.Gtk4/README.md
+++ b/platforms/Linux.Gtk4/README.md
@@ -107,7 +107,7 @@ https://github.com/user-attachments/assets/70f2a910-94b3-437c-945a-6b71223c5cd3
 | Requirement | Version |
 |---|---|
 | .NET SDK | 10.0+ |
-| GTK 4 libraries | 4.x (system package) |
+| GTK 4 libraries | 4.12+ (system package) |
 | WebKitGTK *(Blazor only)* | 6.x (system package) |
 
 ### Install GTK4 & WebKitGTK (Debian / Ubuntu)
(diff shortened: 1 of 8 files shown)


jfversluis (changes requested):
Thanks for adding the fail-fast guard. I think this needs a couple of changes before merging:

1. Blazor startup can bypass the new guard. The sample calls `GtkBlazorWebView.InitializeWebKit()` before `GtkMauiApplication.Run()`, and that path calls `WebKit.Module.Initialize()`, which initializes GirCore/GTK before the new `Gtk.Application.New()` try/catch and version check. In a missing-GTK environment, I still get a raw `DllNotFoundException` from `InitializeWebKit()` before the new friendly message runs. Please share the GTK availability/version guard and invoke it from the WebKit initialization path as well, or otherwise ensure every supported startup path hits it before any GirCore GTK/WebKit initialization.

2. The new GTK 4.12 minimum should be reflected in install/package surfaces. The README still documents `GTK 4 libraries | 4.x`, `.deb` generation still emits `Depends: libgtk-4-1` with no version floor, and the AppImage launcher checks only for `libgtk-4.so` presence. Users on older GTK 4.x distros can install successfully and only fail at launch. Please update the docs and generated dependency/check messages to require GTK 4.12+.

jfversluis (approved):
Re-reviewed the updated changes. The previous blockers are addressed:

- `GtkRuntime.EnsureSupported()` now runs before GirCore GTK initialization from both `GtkMauiApplication.Run()` and `GtkBlazorWebView.InitializeWebKit()`, so the Blazor startup path no longer bypasses the friendly GTK guard.
- The GTK 4.12+ floor is now reflected in the README/docs, generated `.deb` dependency (`libgtk-4-1 (>= 4.12.0)`), and AppImage dependency checks.

I also verified the GTK solution builds and that the Blazor sample now hits the friendly GTK guard before WebKit initialization in a missing-GTK environment.

jfversluis:
Thank you so much for your contribution!

PR #396, opened 2026-07-22
Add native control inspection to DevFlow

morning4coffe-dev:
## Why

DevFlow follows the MAUI visual tree, but real apps also contain platform-owned UI: navigation bars, toolbar buttons, tab overflow, flyouts, dialogs, and other native surfaces. Those controls can be visible and actionable while remaining absent from the logical MAUI tree.

This PR reconciles those two trees without pretending native controls are MAUI elements. Native nodes retain their platform identity, attach to the MAUI object that owns them, and advertise only the actions they support.

## Dependencies and branch relationship

| PR | Purpose |
|---|---|
| [maui-labs#406](https://github.com/dotnet/maui-labs/pull/406) | Splits the framework-neutral agent abstractions from the MAUI backend and adds the plain .NET native agent. This PR is temporarily stacked on that branch. |
| [maui#36821](https://github.com/dotnet/maui/pull/36821) | Emits the shared diagnostics contract and Apple native-element registrations. |
| [maui#36835](https://github.com/dotnet/maui/pull/36835) | Emits Android native-element registrations using the same shared diagnostics contract. |

The diagnostics contract is versioned and string-based, so MAUI does not take a dependency on DevFlow. When #406 merges, this PR will be rebased and retargeted to `main`.

## What changed

- Adds a versioned MAUI native-element diagnostics subscriber.
- Tracks registered native elements with stable IDs, bounded strong retention, Apple native-handle identity, safe stale-peer unregistration, and end-of-walk eviction.
- Produces semantic roles, owner relationships, bounds quality, capture epochs, registry generations, window IDs, and per-element capabilities.
- Adds native discovery and actions for Android, iOS, Mac Catalyst, Windows, macOS AppKit, Linux GTK, and WPF.
- Covers toolbars, navigation/back behavior, Shell tabs and More, flyouts/sidebars, search, dialogs, prompts, action sheets, and system-owned overflow controls where public platform APIs allow it.
- Composes native windows and dialogs into screenshots and returns the most specific native match first during hit-testing.
- Binds actions to the capture that produced an ID and rejects stale or retargeted elements.
- Carries capture metadata through the Driver, CLI, MCP tools, and Web Inspector integration seam.
- Preserves the existing public `AgentClient` API.

Generic reflection over native properties remains read-only. A native element can only be changed through an advertised capability such as `invoke`, `focus`, `set-value`, or `scroll`.

## Integration with #406

This branch deliberately resolves the architectural and runtime concerns raised during the #406 trial merge:

- MAUI-specific implementation moved into `MauiDevFlowAgentService`; neutral routing and contracts remain in `Agent.Abstractions`.
- Only the MAUI backend advertises the v2 capture/native-owner semantics. The plain native backend advertises its supported v1 subset.
- `ElementInfo` uses declared capabilities first and role inference as fallback, with matching behavior in the Driver model.
- The MAUI diagnostics registry was renamed to `RegisteredNativeElementRegistry` to distinguish it from #406's native-tree registry.
- Recently observed Apple peers are held strongly under a bounded policy so IDs survive managed-peer collection without allowing recycled native addresses to retarget.
- Successful physical UI operations invalidate capture leases, preserving stale-capture rejection after the service split.

## Testing

- DevFlow unit tests: 387 passed
- CLI unit tests: 802 passed
- WPF handler tests: 35 passed
- Built successfully: native Android agent; MAUI Android, iOS, Mac Catalyst, and macOS agents; GTK and WPF agents; AppKit backend
- Android browser acceptance: primary/overflow toolbar actions, dynamic registration/removal, alert Accept/Cancel, screenshots, native properties, hit-testing, and stale IDs
- Windows browser acceptance: toolbar actions, ContentDialog discovery and Accept/Cancel, native-ID invocation, screenshot composition, property rejection, hit-testing, cleanup, and stale IDs
- Apple follow-up testing covered iOS, Mac Catalyst, and macOS behavior

## Platform limits

- AppCompat and UIKit do not expose stable public automation objects for every rendered overflow-menu row.
- The UIKit overflow launcher is invokable through public API on iOS/Mac Catalyst 17.4 and later; earlier versions remain select-only.
- More rows are inspectable only while the system More UI is realized.


Redth:
Heads-up from #406 (native/non-MAUI DevFlow agent), which lands in a lot of the same files. I ran a trial merge and read through this branch to work out how the two interact. Short version: the *features* are complementary, but the *structure* collides, and there is one hazard that merges clean and only shows up at runtime.

## The two efforts don't overlap in scope

Worth stating explicitly because "native" means different things in the two PRs:

- **This PR**: a MAUI app has two trees, and platform-owned chrome (nav bars, toolbars, dialogs, More) is missing from the logical one. You reconcile them.
- **#406**: there is no MAUI at all, so there's only one tree. The walker already sees toolbars and nav bars because it *is* walking the platform tree.

So the problem this PR solves genuinely cannot occur in #406's world, and vice versa. No duplicated work, which is good news.

## Trial merge: 6 conflicts, one severe

`git merge-tree` against #406 HEAD. Rename detection handled most of it -- `AgentHttpServer.cs` and `Css/CssSelectorEngine.cs` follow cleanly into `Agent.Abstractions`.

The severe one is **`Agent.Core/DevFlowAgentService.cs` (modify/delete)**. #406 splits that file into four:

| Destination | Lines | Contains |
|---|---|---|
| `Agent.Abstractions/DevFlowAgentService.cs` | 727 | framework-agnostic base |
| `Agent.Abstractions/DevFlowAgentService.Handlers.cs` | 2813 | framework-agnostic handlers |
| `Agent.Abstractions/DevFlowAgentService.Invoke.cs` | 515 | invoke surface |
| `Agent.Core/MauiDevFlowAgentService.cs` | 3452 | MAUI-specific |

This PR adds ~1738 lines to the original. A naive merge leaves the old file in the tree and duplicates **16 types**, including `DevFlowAgentService` itself. That's at least a loud compile failure rather than a silent one.

Four files this PR edits no longer exist on #406: `AgentHttpServer.cs`, `Css/CssSelectorEngine.cs`, `ElementInfo.cs`, `DevFlowAgentService.cs`.

## The one that worries me: capabilities merge clean and break at runtime

This PR bumps `ui.query`, `ui.hit-test`, `ui.actions` and `ui.screenshot` to v2, adding features like `native-owner`, `capture-epoch`, `registry-generation` and `stale-capture-rejection`.

On this branch that lives in `Agent.Core`, which is MAUI-only, so it is correct as written. **#406 moves that capability block into `Agent.Abstractions`, which is shared by the MAUI agent and the plain .NET agent.** After a merge, a non-MAUI agent would advertise v2 features that are derived from MAUI's diagnostic registrations and that it cannot produce. No conflict marker, no compile error, just clients assuming v2 semantics against an agent that can't honor them.

The fix is small and the machinery already exists -- `Capability(version, supported, features, reason)` plus a `PopulateCapabilities` override, so the native backend can advertise the subset it actually supports. It just has to be done deliberately. Flagging it because it isn't visible from either branch alone.

## ElementInfo traits: complementary, but the conflict markers hide that

The one content conflict in `ElementInfo` is more interesting than it looks. Both PRs teach it to derive traits, from different sources:

- This PR: from declared `Capabilities` (`invoke`, `set-value`, `focus`, `scroll`) -- precise, but requires something to publish them.
- #406: from `role` inferred off the native type (`NSButton` -> `button` -> interactive) -- works when nothing publishes capabilities, which is always true in a plain .NET app.

The right resolution is capabilities first with role inference as fallback, not either side wholesale. Taking this PR's side wholesale would silently drop trait inference for every non-MAUI app; taking #406's would lose the precision. Also noticed you dropped `slider` from the interactive role list, presumably because `set-value` covers it -- that's a deliberate difference that a mechanical resolution would lose either way.

## Possibly worth checking in `NativeElementRegistrationRegistry`

This is a hypothesis from reading, not something I've reproduced against your branch, but it's specific enough to be worth a look because I just spent a while fixing the same class of bug in #406.

Identity is minted as `native:registered:{Guid.NewGuid()}` and stored in a `ConditionalWeakTable` keyed on the **managed peer** of the native element, with `WeakReference<object>` held for the element itself.

On Apple, the managed peer for a framework-owned view is recreated on each marshal and is collectable while the native object stays alive. If that happens:

1. the CWT entry disappears, so a re-registration mints a **fresh GUID**, and
2. the old registration's weak ref is dead, so `TryGet` prunes it and returns false.

The old ID is then permanently unresolvable while the control is still on screen, and because the ID is a random GUID there is no path back to it. Your "stale ID" handling would report this as working as intended.

This only bites where MAUI holds no strong reference to the native element -- fine for handler-owned platform views, less obviously fine for the system-owned controls this PR specifically targets (UIKit overflow launcher, nav back button, More rows). It's Apple-specific; Android peers are pinned by the runtime's native/managed mapping.

For what it's worth this is exactly what broke #406's `ios (native)` axis. My fix was to hold recently walked views strongly under a bounded LRU, which makes the native address non-recyclable by construction and removes the liveness ambiguity entirely. Happy to walk through it if useful -- the failure was intermittent and depended on simulator GC timing, so it passed locally and only showed up in CI.

## Suggested order, and the honest cost

I'd suggest #406 lands first and this PR rebases onto the split -- not because of precedence, but because of who is qualified to make the call. Re-targeting the ~1738 lines means deciding, hunk by hunk, MAUI-specific vs framework-agnostic. You know that answer for this code; most of it consumes MAUI diagnostics and would move into `MauiDevFlowAgentService.cs` largely wholesale. If this PR lands first, I'd be redoing the same split over 1738 lines of your code and guessing at the boundary with much less context.

That said, the cost lands on you, and 74 files / 15.7k lines is not a cheap rebase. If you'd rather go the other way I'm fine with it -- I'd just want to pair on the MAUI/abstraction boundary rather than guess.

One small thing regardless of order: `NativeElementRegistrationRegistry` (here) and `NativeElementRegistry` (#406, in `Agent.Native`) mean quite different things. Renaming one would save future maintainers some confusion.

Happy to do the capability split ahead of time so that hazard is closed before either merge -- just say the word.

morning4coffe-dev:
Thanks for the detailed #406 integration analysis. I rebased this work onto #406's split and resolved the issues you called out:

- moved the MAUI-specific implementation into `MauiDevFlowAgentService` while keeping neutral contracts/routing in `Agent.Abstractions`
- kept v2 native/capture capabilities MAUI-only; the plain native backend advertises its actual v1 subset
- resolved traits as capabilities-first with role inference fallback, including the Driver model
- renamed the MAUI diagnostics registry to `RegisteredNativeElementRegistry`
- replaced weak Apple peer identity with bounded strong retention keyed by native handle
- added stale-peer unregister protection and end-of-complete-walk eviction
- restored capture invalidation for successful physical UI operations after the service split

The companion MAUI producer work is now split into [dotnet/maui#36821](https://github.com/dotnet/maui/pull/36821) for Apple/shared diagnostics and [dotnet/maui#36835](https://github.com/dotnet/maui/pull/36835) for Android. Both still use the same v1 event names and payload consumed here.

Local QA is green: 387 DevFlow tests, 802 CLI tests, 35 WPF handler tests, and all affected Android/Apple/GTK/WPF/AppKit builds. This PR is intentionally stacked on #406 until it lands, then it will be rebased and retargeted to `main`.

morning4coffe-dev:
@Redth, the #406 integration changes are now pushed and this PR is stacked directly on `redth-native-devflow-agent`. The detailed resolution and QA summary are in the comment above; a formal review request was not permitted from the fork, so tagging you here.

PR #397, opened 2026-07-23
Add shared DevFlow Inspector for browser, VS Code, and Copilot Canvas

morning4coffe-dev:
## Summary

Adds a broker-hosted DevFlow Inspector shared by the browser, VS Code, and GitHub Copilot Canvas.

- coordinates concurrent CLI, MCP, browser, VS Code, and Canvas callers with an app-scoped mutation lease and process-specific broker registration
- adds screenshot and visual-tree inspection, a searchable tree, runtime-owned property editors, click-to-XAML navigation, and guarded write-back for existing literal XAML attributes
- adds broker-owned workflow recording, Markdown persistence/replay, and Inspector data views for logs, network, preferences, device and sensors, files, WebViews, and native alerts
- adds a typed Node/TypeScript client, a VS Code extension, a project-scoped Canvas extension, and CI coverage for the host packages
- stabilizes Apple-platform broker connectivity and Blazor WebView lifecycle/script loading, including macOS AppKit sample wiring

## API and protocol

- adds `GET /api/v1/ui/elements/{id}/properties` and its property-descriptor schema
- adds process identity to agent status/registration and additive mutation-lease/recording APIs in `AgentClient`
- adds platform alert-driver support and a macOS `PlatformViewDisconnected` lifecycle event

## Validation

- `Microsoft.Maui.DevFlow.Tests`: 477 passed
- `Microsoft.Maui.Cli.UnitTests`: 760 passed
- `Microsoft.Maui.DevFlow.Inspector.Tests`: builds successfully
- `src/DevFlow/js`: tests, build, and VSIX packaging pass
- Canvas extension: tests and recorder self-test pass
- live Inspector suite: 60 passed on Windows and 60 passed on Android
- `DevFlow.Sample.MacOS`: cross-builds successfully


.github/extensions/maui-devflow-canvas/extension.mjs
```diff
@@ -0,0 +1,1075 @@
...
+    store._recordingStatus = r;
+    store._emit();
+    return r;
+  }
+  case "record.list":       return st.recorder.list(store);
+  case "replay": {
+    const status = await store.device.recordingStatus();
+    if (status.recording) return { ok: false, error: "Stop or cancel the active recording before replaying a test." };
+    return replayTest(store, {
+      file: body.file,
```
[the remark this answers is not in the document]
morning4coffe-dev: Fixed in 539f6240. Replay paths supplied through both `/control` and the `replay_test` capability now flow through the same confinement in `replay.mjs`: paths are resolved beneath the recorder's `maui-tests` output root, lexical escapes and sibling-prefix paths are rejected, and canonical-path checks reject symlink/junction escapes. Scenario-name replay remains confined through its slugged path under the same root.

Added regression coverage for valid in-root files, traversal/sibling escapes, and canonical escapes through links.

src/Cli/Microsoft.Maui.Cli/DevFlow/Broker/BrokerServer.cs
```diff
@@ -642,8 +1016,38 @@ private async Task HandleInspectorRoute(HttpListenerContext context, string path
...
+            context.Response.StatusCode = 503;
+            context.Response.ContentType = "text/plain";
+            var msg = Encoding.UTF8.GetBytes("Agent reconnected; reload Inspector state");
+            await context.Response.OutputStream.WriteAsync(msg);
+            context.Response.Close();
+            return;
+        }
+
         // Proxy the request through the inspector's route handler
         await inspector.HandleBrokerRequestAsync(context, subPath);
```
jfversluis, line 1025: This call runs while holding the per-agent `routeGate`, so every non-WebSocket Inspector request is serialized for the full duration of the proxied request. A flow replay executes inline for up to two minutes; during that time state reads, screenshots, heartbeat/status requests, and disconnect cleanup all queue behind this semaphore. It also prevents concurrent mutations from reaching `RouteAsync`, making the new replay-in-progress `409` path unreachable in the primary broker-hosted mode. Please limit the gate to resolving/validating the connection and Inspector instance, then dispatch outside it (or use shared/exclusive connection lifetime coordination), and cover a second request arriving while replay is in flight.

src/DevFlow/js/devflow-client/src/http.ts
```diff
@@ -0,0 +1,126 @@
...
+          Accept: "application/json",
+          ...headers,
+          ...(hostHeader ? { Host: hostHeader } : {}),
+          ...(data ? { "Content-Type": "application/json", "Content-Length": data.length } : {}),
+        },
+      },
+      (res) => {
+        const chunks: Buffer[] = [];
+        res.on("data", (c: Buffer) => chunks.push(c));
+        res.on("end", () => {
```
jfversluis, line 69: The promise only succeeds on `res.end`, while failures are observed only on `req.error` or the request timeout. If the server sends headers plus a partial body and then resets the socket, Node emits `aborted`/`error` on `res` and `close` on `req`; `end` and `req.error` do not fire, and the destroyed socket no longer reaches the timeout. This leaves the promise pending forever and wedges the client's reconnect/resolution mutex. Please add single-settle handling for response abort/error (and request close as appropriate), returning a connection error that `isConnError` recognizes, plus a reset-mid-response test.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Abstractions/AgentHttpServer.cs
```diff
@@ -264,22 +323,79 @@ private async Task<HttpResponse> RouteRequestAsync(HttpRequest request)
...
+        RouteHandler route,
+        HttpRequest request,
+        CancellationToken ct)
+    {
+        var requiresMutationLease = route.RequiresMutationLease &&
+            !(route.MutationLeaseExemption?.Invoke(request) ?? false);
+        var coordinatesLease = route.RequiresMutationLease ||
+            request.Path.Equals("/api/v1/agent/lease", StringComparison.OrdinalIgnoreCase);
+        if (coordinatesLease)
+            await _mutationAdmissionGate.WaitAsync(ct).ConfigureAwait(false);
```
jfversluis: Including `/api/v1/agent/lease` in this admission gate means a forced lease takeover queues behind the active mutation it is intended to preempt. If the UI dispatcher is blocked, the mutation can hold this gate indefinitely, so takeover and all later mutations accumulate even after client HTTP timeouts. Please keep lease-control requests outside the mutation admission semaphore—the lease coordinator already provides its own synchronization—and bound admission waits so a stuck mutation returns the existing retryable busy response rather than wedging the server.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Core/MauiDevFlowAgentService.MutationRecording.cs
```diff
@@ -0,0 +1,207 @@
...
+
+        var avoidText = action == "fill" ||
+            (action == "setProperty" && string.Equals(name, "Text", StringComparison.OrdinalIgnoreCase));
+        var automationId = request.MutationTargetAutomationId ?? target?.AutomationId;
+        var route = await DispatchAsync(() => GetCurrentRouteLocation());
+        return new MutationObservation
+        {
+            Action = action,
+            AutomationId = automationId,
+            Text = !avoidText && string.IsNullOrWhiteSpace(automationId) ? target?.Text : null,
```
jfversluis: `MutationObserver` runs after the route handler, so this `target` comes from the post-mutation tree. For an element without an `AutomationId` that changes its own text—such as the default MAUI counter button—the recording stores `Clicked 1 time` as the replay selector. Replaying from the initial state cannot resolve that selector. Please capture fallback identity (`Text`/`Type`) alongside `MutationTargetAutomationId` before the mutation and use that snapshot for the selector; the post-mutation read should remain only for the resulting value.

.github/extensions/maui-devflow-canvas/store.mjs
```diff
@@ -0,0 +1,1067 @@
...
+    }, 120);
+  }
+
+  async _syncNow() {
+    // Never contend with an in-flight refresh/mutation or another sync.
+    if (this.state.busy || this._polling) return;
+    const agentEpoch = this._agentEpoch;
+    this._polling = true;
+    try {
+      const [t, theme] = await Promise.all([this.device.getRoots(0), this.device.themeGet()]);
```
jfversluis, line 650: The entry guard does not prevent this in-flight sync from becoming stale. A sync can fetch pre-mutation tree A, then a mutation-triggered `refresh()` applies post-mutation tree B, after which this continuation resumes and applies A over B with a newer revision. The panel regresses to stale state until the next poll. Please capture a refresh generation before this await and discard the result if it changed, or serialize `_syncNow` through the same refresh chain used by `refresh()`.

jfversluis (changes requested):
Requesting changes for five correctness issues in the new concurrency, transport, recording, and live-sync paths. The two hangs can wedge primary Inspector clients, while the recording and synchronization issues produce incorrect user-visible behavior.

src/Cli/Microsoft.Maui.Cli/DevFlow/Broker/BrokerServer.cs
morning4coffe-dev: Good catch — the gate was doing double duty. Replaced the per-agent mutex with a shared/exclusive lifetime gate (`AgentLifetimeGate`): Inspector requests take the **shared** side, while a same-ID reconnect and disconnect cleanup take the **exclusive** side. That keeps the lifetime guarantee you were relying on (nothing can swap or dispose the agent underneath an in-flight request) while letting requests run concurrently.

Measured on a live app with an 8-step replay in flight:

| | before | after |
|---|---|---|
| `GET /api/state` | blocked **8.9 s** (returned only after the replay finished) | **178-252 ms**, served while the replay was still running |
| concurrent mutation | never reached `RouteAsync` | `409` in 3 ms - *"A replay is in progress"* |

So the replay-in-progress 409 is reachable in broker-hosted mode now. Covered by `AgentLifetimeGateTests` (concurrency, exclusion, cancellation, double-dispose) plus `Mutation_ArrivingWhileReplayIsInFlight_ReturnsConflict`, which parks a replay inside the agent and asserts the second request gets 409 rather than blocking.

src/DevFlow/js/devflow-client/src/http.ts
morning4coffe-dev: Confirmed, and it turned out to be slightly different from (and a bit worse than) described. I reproduced all three teardowns against a real socket on Node v24:

| server tears down mid-body | events seen | settles? |
|---|---|---|
| `socket.end()` (graceful FIN) | `res.aborted`, `req.close`, `res.error` | **never** |
| `socket.destroy()` | same | **never** |
| `resetAndDestroy()` (true RST) | `req.error:ECONNRESET` first | yes |

So a literal RST *does* settle via `req.error` - the promise only hangs on FIN/`destroy()`, which is the common case (agent killed mid-response). The request timeout never fires either, because the socket is already gone.

Fixed with a single-settle guard: every terminal path goes through one `settle()`, and the response side is now handled (`aborted`, `error`, and `close` while `!res.complete`), resolving `{ ok:false, status:0, error:"ECONNRESET" }` which `isConnError` recognises. `req.close` is only a fallback for "socket died before any response" - settling on it unconditionally would truncate close-delimited bodies, since Node emits `req.close` *before* `res.end` there.

Tests cover all three teardowns plus a complete response and a close-delimited body. I verified they fail without the fix (the two hanging cases time out at 12 s).

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Abstractions/AgentHttpServer.cs
morning4coffe-dev: Both parts done.

**Lease control is out of the gate.** `coordinatesLease` is gone - only `route.RequiresMutationLease` admits to the semaphore now, so `/api/v1/agent/lease` never queues behind the mutation it exists to preempt. As you noted, `MutationLeaseCoordinator` already serialises itself under its own lock, so the extra gating bought nothing.

**Admission waits are bounded.** `WaitAsync(MutationAdmissionTimeout, ct)` (10 s, internally settable for tests) instead of an unbounded wait; on timeout the request gets a retryable `503` with `reason:"busy"` and `Retry-After: 1` rather than accumulating behind a wedged mutation.

`ForcedLeaseTakeover_WaitsForAdmittedMutation` codified the old behaviour, so I inverted it to `ForcedLeaseTakeover_IsNotBlockedByAnAdmittedMutation` - it now asserts the takeover completes *while* the mutation is still parked in its handler. Added `QueuedMutation_ReturnsRetryableBusy_WhenTheAdmittedOneIsWedged` for the 503 path.

Live check on a real app: 25 mutations replayed back-to-back while hammering the agent's lease endpoint - 17 calls, all `200`, max 41 ms, no spurious 503s.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Core/MauiDevFlowAgentService.MutationRecording.cs
morning4coffe-dev: Exactly right, including the counter-button example. `CaptureMutationTarget` now snapshots `Text` and `Type` alongside the AutomationId *before* the handler runs, and the recorded selector uses that snapshot; the post-mutation walk is only used for the resulting **value**. It's also called from the fill and setProperty handlers now, not just tap.

The text comes from a new shared `VisualTreeWalker.ExtractText`, which the tree walker itself also uses - so a recorded text selector is byte-identical to what a query or replay searches for (password redaction included).

Verified end-to-end against a real app, with a no-AutomationId button whose `Clicked` handler rewrites its own text:

- button read `"Clicked 1 time"` by observation time, recorded selector was `"Click me"`
- replayed from a fresh app start: **passes**
- same flow with the old post-mutation selector: **fails** with *"tap target could not be resolved"* - the failure you predicted

Unit coverage in `FlowRecordingTests`: the snapshot itself, password redaction, and that `BuildSelector` produces the pre-mutation text.

.github/extensions/maui-devflow-canvas/store.mjs
morning4coffe-dev: Took the refresh-generation approach. Added `_treeGen`, bumped on every authoritative `_applyRoots`; `_syncNow` captures it before the fetch and drops its result if it moved - or if a refresh is in flight (`state.busy`), since that one is authoritative and about to land.

The `rev` detail you called out is what made this nasty: the browser only discards *older* revisions, so the stale tree arriving under a higher `rev` was accepted and stuck until the next poll.

The regression test parks a sync on `getRoots`, lets a `refresh()` publish the newer tree, then releases the sync and asserts the tree is still the post-mutation one and that no new revision was emitted. It fails without the fix (*"stale sync must not regress the tree"*).

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Abstractions/AgentHttpServer.cs
```diff
@@ -328,22 +390,103 @@ private async Task<HttpResponse> RouteRequestAsync(HttpRequest request)
...
+            return await RunRouteAsync(route, request, requiresMutationLease).ConfigureAwait(false);
+
+        if (!await _mutationAdmissionGate.WaitAsync(MutationAdmissionTimeout, ct).ConfigureAwait(false))
+        {
+            // Bounded, not indefinite: a wedged mutation must not accumulate every later
+            // request behind it until each client times out on its own.
+            var busy = HttpResponse.Error(
+                "The app is still applying another mutation. Retry shortly.",
+                statusCode: 503,
+                reason: "busy");
```
jfversluis: This response says to retry and includes `Retry-After`, but the client stack treats it as terminal. `AgentClient` only marks an error retryable when `details.retryable` is true or the reason is `stale-capture-epoch`/`ui-mutation-busy`; the new `busy` reason matches neither. Consequently Inspector replay stops, MCP throws without retry guidance, and CLI output reports `retryable: false`. Please use the established `ui-mutation-busy` envelope with `details: new { retryable = true }` (preferred), or otherwise teach `AgentClient` that `busy` is retryable. A test should assert the response survives `AgentClient` decoding with `Retryable == true`, not only that the raw 503 contains `Retry-After`.

jfversluis (changes requested):
The five original findings are fixed and the added regression coverage addresses the reported scenarios. One issue remains in the bounded mutation-admission fix: its busy response is not recognized as retryable by the first-party clients.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Abstractions/AgentHttpServer.cs
morning4coffe-dev: Adjusted the admission-gate response to use the established `ui-mutation-busy` envelope with `details: { retryable = true }`, so it now survives `AgentClient` decoding as `Retryable == true` — Inspector replay retries instead of treating it as terminal, MCP emits retry guidance, and the CLI reports `retryable: true`.

Kept the `503` + `Retry-After`: teaching `AgentClient` that any 503 is retryable would have wrongly retried the genuinely terminal `reason: "broker"` 503 that `/api/v1/agent/recording` returns from the same gated path.

Added `QueuedMutation_BusyResponse_SurvivesAgentClientDecodingAsRetryable`, which wedges a real `AgentHttpServer` and asserts the decoded `ActionResult` from a real `AgentClient.TapResultAsync` — not just the raw 503 headers.

Thanks!

jfversluis (approved):
Re-reviewed the incremental fix. The mutation-admission timeout now returns the established `ui-mutation-busy` envelope with `details.retryable: true`, and the new integration test verifies the response decodes through a real `AgentClient` as retryable. All previously requested changes are addressed.

PR #406, opened 2026-07-25
Run the DevFlow agent in plain .NET apps, with no MAUI references

Redth:
Makes the DevFlow agent work in plain .NET Android / iOS / Mac Catalyst / macOS apps that reference no MAUI libraries at all — same HTTP API, same CLI, same MCP tools.

## The problem

`Agent.Core` targets `net10.0` but hard-references `Microsoft.Maui.Controls` and `Microsoft.Maui.Essentials`, so hosting the agent meant being a MAUI app. Auditing its ~11,000 lines showed the coupling is very unevenly distributed: roughly 2,500 lines were *already* framework-neutral (HTTP server, CSS engine, network capture, profiler, extensions, WebView CDP registry, actions/invoke), about 3,200 are MAUI UI, and about 1,400 are Essentials. Most of DevFlow never needed MAUI — it was just on the wrong side of an assembly boundary.

## What changed

**`Microsoft.Maui.DevFlow.Agent.Abstractions`** (new, `net10.0`, zero MAUI references) takes the neutral core, including `DevFlowAgentService` itself. Routing is framework-neutral and points at `protected virtual` handlers that default to a `501 not_supported` envelope; backends override what they can do. `Agent.Core` keeps its PackageId and becomes the MAUI backend.

The seam is virtual handlers rather than the `IUiBackend` interface originally sketched, because partial classes cannot span assemblies and `DevFlowAgentService` had to stay in one piece. Orchestration policy — retry, waits, batch sequencing, error shaping, CSS filtering over `ElementInfo` — stays neutral; only primitives that touch a UI framework are overridden.

**`Microsoft.Maui.DevFlow.Agent.Native`** (new, multi-targeted android/ios/maccatalyst/macos) implements Android views, UIKit and AppKit backends: roots, traversal, ids, text, bounds, tap, fill/clear, focus, key, gesture, scroll, screenshot, property get/set. **`.Agent.Native.Essentials`** is an optional add-on — without it, device/storage/sensor/theme groups report `supported: false` rather than disappearing.

**Protocol.** `agent.framework` is now dynamic (`maui` | `native`) and `agent.uiFramework` is new (`maui-controls` | `android-views` | `uikit` | `appkit` | `gtk` | `wpf`). Capabilities gained `supported` + `reason` so clients can discover what a host *cannot* do; unsupported endpoints return a uniform `501 { error: "not_supported", capability, reason }`. `BrokerRegistration` carries both fields, so `maui devflow list` shows native apps alongside MAUI ones.

**Four sample apps** under `samples/DevFlow.Sample.Native/`, sharing the MAUI sample's automation ids so the same assertions run against both.

## Testing

The integration suite gained a `DEVFLOW_TEST_FRAMEWORK` axis (`maui` default | `native`) and a `MacOSFixture`, which is the first automated driver the AppKit backend has ever had. MAUI-only tests (Shell navigate, theme, storage, sensors) are tagged `[Trait("framework", "maui")]` and filtered out of native runs.

| Axis | Result |
|---|---|
| Native / macOS AppKit | 56/56 |
| Native / Mac Catalyst | 56/56 |
| MAUI / Mac Catalyst | 115/116 |
| Unit | 245/245 |

The single MAUI failure is `WebViewTests.Screenshot_WebView_ReturnsImage`, a pre-existing flake: it fails in-suite, passes in isolation, and was reproduced at unmodified `HEAD` across five runs before any of this work. Deliberately not fixed here to keep the change scoped — worth its own issue.

Android and AppKit were also driven by hand against live apps, since the axis reaches UIKit through Mac Catalyst but nothing else. The decisive Android check was scrolling a `LinearLayout`: its own screen Y moved 269.7 → 369.5, which only happens if the enclosing `ScrollView` was resolved and scrolled — the old `View.ScrollBy` path shifts children and leaves the layout in place.

Two bugs worth calling out, both found by running things rather than reading them:

- **AppKit published only native property names.** `GetProperties` returned `StringValue`/`Title`/`Hidden` while `TrySetProperty` already accepted `Text`/`IsVisible`/`IsEnabled`/`IsChecked`, and UIKit published both — so on AppKit you could write a property you could not read back. Mac Catalyst shares the UIKit backend, so its green runs had been hiding the asymmetry. This surfaced on `MacOSFixture`'s very first run.
- **Native scroll didn't walk ancestors.** MAUI resolves scroll targets via `FindAncestor<ItemsView>`, so naming a list's child scrolls the list; native only scrolled when the target was itself a scroll view.

Also fixed: `framework` is a `workflow_dispatch`-only input, so label-triggered PR runs silently fell back to `maui` and would never have exercised the native axis — and that label is the suite's only automatic trigger. PR runs now default to both.

## Breaking changes

Types moved from `Microsoft.Maui.DevFlow.Agent.Core` to `Microsoft.Maui.DevFlow.Agent.Abstractions` (22 files: `AgentHttpServer`, `AgentOptions`, `AgentExtension`, `BrokerRegistration`, `ElementInfo` + DTOs, `FileStoragePathResolver`, `DevFlowAction*`, `Css/`, `Network/`, `Profiling/`, `DevFlowAgentService`). Namespaces are unchanged, so source generally still compiles, but this is **binary breaking** — acceptable at `0.1.0-preview`. Some `protected` members widened to `public` to cross the assembly boundary.

`Agent`, `Agent.Gtk` and `Agent.WPF` compile unchanged: every `protected virtual` hook survives verbatim on `MauiDevFlowAgentService`, which was the explicit constraint on the split.


src/DevFlow/Microsoft.Maui.DevFlow.Agent.Native/NativeElementRegistry.cs
```diff
@@ -0,0 +1,112 @@
...
+    public string Register(object view, string? parentId)
+    {
+        lock (_gate)
+        {
+            var stableKey = GetStableKey(view);
+            string id;
+
+            if (stableKey != null)
+            {
+                if (!_idsByStableKey.TryGetValue(stableKey, out id!))
```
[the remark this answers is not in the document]
Redth: Agreed on both counts, fixed. `BeginWalk` now prunes `_idsByStableKey` (and its gate also fires when that map grows, since it can outlive `_viewsById`), and `Register` validates a stable-key hit: the id is only reused while it still resolves to a live view sitting at that same handle.

Worth flagging one thing your analysis surfaced that pruning alone does not solve. A recycled handle and a legitimate managed-peer recreation are the *same* observable signal -- old peer dead, fresh registration at the same handle -- so they cannot be told apart from the handle alone, and the prune is gated behind the 512 cap so it cannot be relied on to run in time. I took the safe direction and retired the id rather than reviving it, which means `ResolveView`'s re-walk no longer recovers ids after peer collection (its comment no longer claims it does). Filed #413 to get that durability back properly via ObjC zeroing weak references, which distinguish the two cases exactly.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Native/NativeUi.Android.cs
```diff
@@ -0,0 +1,645 @@
...
+    public static string UiFrameworkName => "android-views";
+
+    public static string DeviceTypeName =>
+        Build.Fingerprint?.Contains("generic", StringComparison.OrdinalIgnoreCase) == true ? "virtual" : "physical";
+
+    /// <summary>
+    /// The activity the agent walks. Set by <see cref="DevFlowAgent"/> at bootstrap and refreshed
+    /// on every resume so the tree always reflects the foreground activity.
+    /// </summary>
+    public static Activity? CurrentActivity { get; set; }
```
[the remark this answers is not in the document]
Redth: Applied -- `CurrentActivity` now goes through `Volatile.Read`/`Volatile.Write`, since a rebind can otherwise sit in a store buffer on weakly ordered ARM and leave workers serving the previous foreground activity.

The `NullReferenceException` itself is not reachable though: every read site is either null-guarded or reads once into a local and degrades (`GetRoots` returns empty, `DisplayDensity` returns 1.0), and both writers reject null and run before the HTTP listener starts, so a worker never observes null. Worth doing for the staleness window rather than the NPE.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Native/DevFlowAgent.Android.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// Starts the agent and binds it to <paramref name="activity"/>.
+    /// </summary>
+    /// <param name="activity">The activity whose window the agent walks.</param>
+    /// <param name="options">Optional agent configuration.</param>
+    /// <returns>The running agent service.</returns>
+    public static NativeDevFlowAgentService StartDevFlowAgent(this Activity activity, AgentOptions? options = null)
+    {
+        ArgumentNullException.ThrowIfNull(activity);
+
+        NativeUi.CurrentActivity = activity;
```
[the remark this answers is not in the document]
Redth: Took your documentation fallback rather than the weak reference. The hold is bounded rather than accumulating: both writers overwrite the static, so at most one destroyed activity is retained and the next bind releases it.

`WeakReference<Activity>` would actively hurt here -- if a backgrounded activity is collected mid-session the getter resolves null and the agent silently returns an empty tree and null app info, so the tool goes dark exactly when you are trying to diagnose something. I documented the intentional strong hold on `CurrentActivity` instead.

src/DevFlow/Microsoft.Maui.DevFlow.Driver/AgentClient.cs
```diff
@@ -926,6 +942,47 @@ private async Task<T> SendWithTransientRetriesAsync<T>(HttpMethod method, Func<T
...
+                reason = r.GetString();
+        }
+        catch (JsonException)
+        {
+            // Non-JSON 501 — leave it to the caller's existing error handling.
+            return;
+        }
+
+        // The caller never receives the message, so release it here.
+        response.Dispose();
```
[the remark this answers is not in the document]
Redth: Declining this one -- it is not a double dispose, and removing the explicit `Dispose` would introduce a leak. `ThrowIfNotSupportedAsync` has exactly one caller, `SendWithTransientRetriesAsync`, which holds the message in a plain local (`var result = await send()`) and throws before `return result`. The transitive callers do use `using var response = await ...`, but on the throw path the await rethrows before `response` is ever bound, so that `using` never takes ownership.

Confirmed with a small repro: without the explicit dispose the 501 response gets 0 disposes, with it exactly 1. Leaving the code and the comment as-is.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Abstractions/ElementInfo.cs
```diff
@@ -157,6 +158,41 @@ public ElementNativeViewInfo? NativeView
...
+
+        // AppKit
+        "NSWindow" => "window",
+        "NSSecureTextField" or "NSSearchField" or "NSTextView" => "textbox",
+        "NSImageView" => "image",
+        "NSTableView" or "NSOutlineView" or "NSCollectionView" => "list",
+        "NSSlider" or "NSStepper" => "slider",
+        "NSProgressIndicator" => "progressbar",
+        "NSSegmentedControl" => "tablist",
+        "NSTextField" => "text",
```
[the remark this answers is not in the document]
Redth: Good catch, fixed. `NSButton` now maps to `"button"`, and I added `NSSwitch` to `"switch"` for parity with `UISwitch`. Both now pick up the `interactive` and `focusable` traits, matching UIKit and Android. Added unit tests covering the new mappings plus cross-platform button-role parity.

Left out the checkbox and radio part deliberately: AppKit has no distinct classes for those, they are `NSButton` with a different `ButtonType`, and the macOS sample shows exactly that with `SetButtonType(NSButtonType.Switch)` on an `NSButton`. Since the backend reports the exact runtime class name, a type-name switch can only resolve them to `"button"` -- separating them would need the backend to emit a refined type string, which felt out of scope here. Also skipped `NSPopUpButton`: it is a real subclass so it would need its own arm, but it is a menu control and there is no fitting role in the current vocabulary, so `null` beats mislabeling it a button.

samples/DevFlow.Sample.Native/MacOS/AppDelegate.cs
```diff
@@ -0,0 +1,196 @@
...
+        Record(null);
+    }
+
+    /// <summary>Pushes model state into the status/count labels. Pass null to only re-read.</summary>
+    private void Record(string? action)
+    {
+        if (action is not null)
+            _model.RecordAction(action);
+
+        _status.StringValue = _model.Status;
```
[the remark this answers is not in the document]
Redth: Checked this empirically and the premise does not hold. .NET for macOS does install a main-thread `SynchronizationContext` under plain `NSApplication.Main()`, so the continuation resumes on the main thread. I put a temporary probe around the `GetPostsButton` handler and ran the sample:

```
[SYNCCTX] launch          ctx=AppKit.AppKitSynchronizationContext isMain=True tid=1
[SYNCCTX] before-await    ctx=AppKit.AppKitSynchronizationContext isMain=True tid=1
[SYNCCTX] after-await     ctx=AppKit.AppKitSynchronizationContext isMain=True tid=1
```

So the `NSTextField` writes in `Record` already run on the main thread, same as the UIKit heads, and macOS is not a weaker case than iOS/Android here. Leaving it as-is rather than adding an `InvokeOnMainThread` that would make the heads inconsistent and model less idiomatic async-UI practice.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Abstractions/IAgentDispatcher.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    /// <inheritdoc />
+    public bool Dispatch(Action action)
+    {
+        _dispatch(action);
+        return true;
+    }
+
+    /// <inheritdoc />
+    public bool DispatchDelayed(TimeSpan delay, Action action)
```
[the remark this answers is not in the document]
Redth: Agreed on the drift risk, addressed slightly differently. Both the default interface method and the `DelegateAgentDispatcher` override now delegate to a single internal `AgentDispatcherDefaults.DispatchDelayed` helper, so there is exactly one copy of the logic.

Kept the concrete override rather than deleting it: a default interface method is only reachable through the interface, and `DelegateAgentDispatcher` is public in a shipping package, so removing it would be a source-breaking change for anyone holding a concrete-typed reference. `MauiAgentDispatcher`'s own override is a genuine adapter over MAUI's `IDispatcher` and is untouched.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Native/NativeElementRegistry.cs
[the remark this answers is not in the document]
Redth: Following up because my earlier fix here was wrong and I want the record to be accurate.

Gating id reuse on the previously bound peer still being alive regressed the `ios (native)` axis. On the Apple backends the managed peer for a framework type is recreated on each marshal and is collectable while the native view lives on, so the id was being retired on a routine GC rather than on an actual handle recycle. I had argued the resulting miss would be "loud and self-correcting"; it is neither, since the client's cached id becomes permanently unresolvable for a view that is still onscreen.

Replaced with strong holding: recently walked views are pinned, bounded at 512 and evicted least-recently-seen, with each evicted id dropping its stable key alongside. That closes the hazard you identified by construction instead of by detection, since a native address cannot be recycled while we still hold the view, and it removes the liveness ambiguity entirely. Guarded by ten new unit tests that drive collection deterministically and were verified to fail against the previous implementation. Thanks for pushing on this one, the underlying concern was correct.

morning4coffe-dev:
Follow-up [#396](https://github.com/dotnet/maui-labs/pull/396) is now stacked directly on this branch. Its MAUI-specific implementation was moved onto the split, the native backend keeps its v1 capability subset, trait fallback is preserved, and the diagnostic registry lifetime/naming concerns are resolved. No #406 code changes are required for the integration.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Native/NativeDevFlowAgentService.cs
```diff
@@ -0,0 +1,690 @@
...
+            });
+
+            return matches
+                .OrderByDescending(match => match.Depth)
+                .ThenBy(match => (match.Element.Bounds?.Width ?? double.MaxValue) * (match.Element.Bounds?.Height ?? double.MaxValue))
+                .Select(match => match.Element)
+                .ToList();
+        });
+
+        return HttpResponse.Json(hits);
```
JanKrivanek: Could we preserve the documented hit-test response envelope here? `InspectorServer` looks for a root-level `elements` property and returns no candidate when it is absent, so returning the bare `hits` array makes click-to-select silently fail against native agents. The OpenAPI contract also requires the `x`, `y`, `window`, `captureEpoch`, `registryGeneration`, and `elements` fields.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Native/NativeUi.AppKit.cs
```diff
@@ -0,0 +1,547 @@
...
+        var type = view.GetType();
+        var window = view.Window;
+
+        // AppKit's origin is bottom-left; DevFlow reports top-left screen coordinates.
+        var inWindow = view.ConvertRectToView(view.Bounds, null);
+        var screenY = window == null
+            ? inWindow.Y
+            : (window.Screen ?? NSScreen.MainScreen)!.Frame.Height
+              - (window.Frame.Y + inWindow.Y + inWindow.Height);
+        var screenX = window == null ? inWindow.X : window.Frame.X + inWindow.X;
```
JanKrivanek: These bounds are screen-absolute, while the native agent advertises `window-logical-coordinates` and `HandleHitTest` compares the requested coordinates directly against them. UIKit currently reports window-relative bounds, so the backends also disagree. On a window moved away from the screen origin, hit testing and Inspector overlays will be offset. Can we normalize every native backend to the advertised coordinate space and populate `WindowBounds` as needed?

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Native.Essentials/EssentialsNativeDevFlowAgentService.cs
```diff
@@ -0,0 +1,104 @@
...
+    public EssentialsNativeDevFlowAgentService(AgentOptions? options = null)
+        : base(options)
+    {
+    }
+
+    /// <inheritdoc />
+    protected override bool IsStorageSupported => true;
+
+    /// <inheritdoc />
+    protected override bool IsDeviceInfoSupported => true;
```
JanKrivanek, line 34: This causes capabilities to advertise `device.info` as supported, including the `app` feature, but this service does not override `HandlePlatformAppInfo`; that route therefore falls through to the base 501 response. Capability-aware clients will be told the operation is available and then fail. Could we either implement app info from the values already exposed by `NativeUi`, or report the supported features more narrowly?

src/DevFlow/Microsoft.Maui.DevFlow.Driver/AgentClient.cs
```diff
@@ -915,7 +1630,15 @@ private async Task<T> SendWithTransientRetriesAsync<T>(HttpMethod method, Func<T
...
             try
             {
-                return await send();
+                var result = await send();
+
+                // A backend that cannot serve an endpoint answers 501 with a not_supported
+                // envelope. Surface that as a typed exception instead of letting it decay into a
+                // null/false that looks identical to a genuine failure.
+                if (result is HttpResponseMessage message)
+                    await ThrowIfNotSupportedAsync(message);
```
JanKrivanek, line 1684: This also changes the contract of the `*ResultAsync` APIs: the shared send path throws `NotSupportedByAgentException` before `ScreenshotResultCoreAsync`, `HitTestResultAsync`, or `SendActionResultAsync` can return their documented structured failure result. The existing catch filters do not include this exception. Can we preserve non-throwing behavior for the result-returning APIs, or explicitly update their public contract and all callers?

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Native.Essentials/EssentialsNativeDevFlowAgentService.cs
Redth: Good catch, thanks. Implemented `HandlePlatformAppInfo` in `EssentialsAgentSupport` (backed by `AppInfo.Current`, mirroring the existing `HandlePlatformDeviceInfo`) and wired it into `EssentialsNativeDevFlowAgentService`, so the "app" feature under `device.info` now actually works instead of falling through to the base 501. Pushed in 1efaa59e.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Native/NativeDevFlowAgentService.cs
Redth: Good catch, agreed. Fixed in e4f1e9a: `HandleHitTest` now wraps the hits in the full envelope (`x`, `y`, `window`, `captureEpoch`, `registryGeneration`, `elements`) instead of returning a bare array, and it now honors the `window` query parameter too (it previously always scanned every window).

For the epoch fields, I kept them honest to what the native agent actually provides rather than borrowing MAUI's semantics: `captureEpoch` is `NativeElementRegistry`'s walk counter (newly exposed as `CurrentWalk`), which increments on every tree walk and satisfies the OpenAPI `minimum: 1` once a walk has run. `registryGeneration` has no independent concept on the native backend (it doesn't advertise MAUI's `stale-capture-rejection` feature), so it stays at the schema's minimum default of 0 rather than faking a second dimension of staleness tracking.

Added `NativeElementRegistryTests.CurrentWalk_*` unit tests plus an integration test (`UiInspectionTests.HitTest_Response_MatchesDocumentedEnvelope`) that parses the raw JSON and asserts the envelope shape. I verified against the actual native macOS AppKit sample (`DEVFLOW_TEST_FRAMEWORK=native`) that the new test fails with `Expected: Object / Actual: Array` on the old code and passes with the fix, plus the full native integration suite (91 tests) still passes.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Native/NativeUi.AppKit.cs
Redth: Good catch, and confirmed by testing: AppKit and Android were both reporting screen-absolute X/Y, while UIKit's `ConvertRectToView(bounds, null)` was already window-relative. Since `ui.hit-test` advertises `window-logical-coordinates` and `HandleHitTest` compares x/y directly against `ElementInfo.Bounds`, the AppKit/Android bounds would drift out of sync with hit testing, screenshots, and Inspector overlays as soon as a window moved off the screen origin, exactly as you described.

Fixed in 50628d7:
- AppKit: flip against the window's own frame height instead of the screen's, so `Describe()` returns top-left, window-relative bounds. Extracted the flip into a small AppKit-free `NativeUi.FlipAppKitWindowBaseToTopLeft` helper so it's unit-testable without a live `NSWindow`.
- Android: swapped `GetLocationOnScreen` for `GetLocationInWindow`, which is already window-relative.
- Updated `NativeViewDescriptor`'s doc comment, which incorrectly said "relative to the screen".

I did not add a `WindowBounds` value for the native-only agent: unlike the MAUI agent (which nests native elements inside a MAUI page and uses `WindowBounds` to give a more precise window-relative bound alongside a page-relative `Bounds`), the native-only agent has no second coordinate space to reconcile - every element's `Bounds` is now directly and only window-relative, so a duplicate `WindowBounds` would just repeat the same value.

I added `NativeUiCoordinatesTests`, which covers the flip algebra across several window positions/sizes, including a window moved well away from `(0, 0)` on a taller screen and one moved to negative screen coordinates (secondary monitor) - the exact regression this fixes. I also attempted a live `NSWindow`/`NSView` integration check on this machine (real macOS), but the installed Xcode (26.6) doesn't match what this .NET for macOS SDK build requires (26.2), so I couldn't produce a runnable app bundle here to exercise it end-to-end; happy to add one if there's an existing harness/CI runner where that's viable.

src/DevFlow/Microsoft.Maui.DevFlow.Driver/AgentClient.cs
Redth: Good catch, this is a real regression and I agree with your read. The shared retry path throws `NotSupportedByAgentException` for the uniform 501 envelope before `ScreenshotResultCoreAsync`, `HitTestResultAsync`, or `SendActionResultAsync` (backing all the `*ResultAsync` action APIs) can return it, silently breaking the documented non-throwing contract for the whole `*ResultAsync` family and the callers that rely on it (`ScreenshotTool`, `HandleProxyTapAsync`, `HitTestAsync`/`ScreenshotAsync`).

Fixed by catching `NotSupportedByAgentException` in those three methods and translating it into a failure result (`Reason = "not_supported"`, `Retryable = false`), consistent with how every other structured failure is reported. Plain convenience APIs like `TapAsync`/`GetTreeAsync` are untouched and keep throwing, since that's their established, already-tested contract.

Added tests proving the regression (`ScreenshotResultAsync`, `HitTestResultAsync`, `TapResultAsync` all previously threw against a 501 not_supported agent, now return structured failures) and proving the convenience-API distinction is preserved (`TapAsync` still throws). Pushed as 44f5dedf.

jamesmontemagno (approved):

src/DevFlow/Shared.Essentials/EssentialsAgentSupport.cs
```diff
@@ -0,0 +1,806 @@
...
+        try
+        {
+            var info = AppInfo.Current;
+            return Task.FromResult(HttpResponse.Json(new
+            {
+                name = info.Name,
+                packageName = info.PackageName,
+                version = info.VersionString,
+                buildNumber = info.BuildString,
+                requestedTheme = info.RequestedTheme.ToString(),
```
JanKrivanek: These AppInfo properties are UIKit-backed on iOS and Mac Catalyst, but this handler runs on the HTTP worker thread. Reading `RequestedTheme` or `RequestedLayoutDirection` there can throw `UIKitThreadAccessException`, turning the newly supported `/api/v1/device/app` route into a platform error while capabilities advertise it as available. The MAUI implementation dispatches the same reads through `MainThread.InvokeOnMainThreadAsync`; could we do that here as well?

JanKrivanek (approved):
Looks good - one more finding before ready to go

src/DevFlow/Shared.Essentials/EssentialsAgentSupport.cs
Redth: Good catch, and confirmed: the iOS/Mac Catalyst `AppInfoImplementation` reads `RequestedTheme` off `UIScreen.MainScreen.TraitCollection` and `RequestedLayoutDirection` off the current `UIWindow`, both UIKit APIs. I wrapped the read in `MainThread.InvokeOnMainThreadAsync`, matching the existing pattern in `HandlePlatformDeviceDisplay` right below it in the same file and the MAUI agent's own `HandlePlatformAppInfo` override, so both consumers of `EssentialsAgentSupport` now dispatch consistently. Pushed in 907fad5. Verified builds for all native Essentials TFMs (android/ios/maccatalyst/macos) and the affected protocol/capability/theme unit tests still pass; I did not add a new unit test since there is no injectable main-thread dispatcher abstraction here yet and the existing `HandlePlatformDeviceDisplay` sibling has no dedicated unit test either, so a device-dependent test would be the only option and that is explicitly what we want to avoid.

JanKrivanek (approved):
Thanks for iterating on improvements

jfversluis (approved):

PR #412, opened 2026-07-27
Add cross-platform layout diagnostics to MAUI DevFlow

morning4coffe-dev:
## Summary

This adds a first-class layout diagnostics pipeline to MAUI DevFlow. It gives developers and agents structured information about UI problems that are difficult to infer from a screenshot alone, including clipped elements, content overflow, text truncation, zero-area controls, overlap, and visual or interaction occlusion.

The feature is intentionally gated behind `AgentOptions.EnableLayoutDiagnostics` and remains off by default while the remaining Apple and GTK runtime matrix is completed.

## Why this is useful

Screenshots show the final pixels, but they do not explain why something is missing or difficult to interact with. An agent cannot reliably tell whether text was ellipsized, content extends beyond a clipped host, a control has no usable area, or another element receives its input.

Layout diagnostics adds that missing structural evidence. Findings include the affected element, source location when available, severity, confidence, actionability, related clippers or occluders, and optional geometry evidence. Unsupported or incomplete coverage is reported explicitly instead of being treated as a clean pass.

## How it works

The agent captures an immutable layout snapshot from the realized MAUI tree and enriches it with platform-specific information from Android, iOS/Mac Catalyst, macOS, WinUI, WPF, GTK, and Blazor WebView collectors. The analysis uses transformed regions, clip chains, native text metrics, hit-test sampling, and a window-wide spatial index while keeping geometric overlap, visual occlusion, and interaction occlusion as separate concepts.

The same versioned result is available through:

- `POST /api/v1/ui/diagnostics/layout`
- the DevFlow Driver API
- `maui devflow ui diagnostics`
- the `maui_layout_diagnostics` MCP tool
- the DevFlow Inspector diagnostics panel

The contract also includes stability sampling, privacy controls, persistent project/user suppressions, compact agent output, source mapping, deterministic suppression keys, rule coverage, and explicit incomplete results for timeouts or opaque subtrees.

## DevFlow Inspector integration

This PR includes the initial Inspector experience: filtering, source-aware findings, polygon and clip overlays, related-element navigation, suppression controls, immutable screenshot frames, and copyable agent payloads. It also publishes finding deltas over the existing Inspector event stream.

A follow-up Inspector workflow can build on this by making diagnostics part of element inspection itself: select an element, see the active layout issue and its clip/occluder, then send that exact finding to an agent with the relevant source and geometry context. This provides a path from observing a visual problem to asking an agent to repair and re-check it without relying only on screenshots.

## Trying it

Enable the feature in the app:

```csharp
builder.AddMauiDevFlowAgent(options =>
{
    options.EnableLayoutDiagnostics = true;
});
```

Then run:

```bash
maui devflow ui diagnostics --profile agent
```

The DevFlow sample includes a dedicated page with clipping, truncation, overflow, intentional overlap, and blocking-overlay fixtures.

## Validation

- 296 DevFlow tests passed
- 755 CLI tests passed
- Windows native and Blazor runtime diagnostics: 4/4 passed
- Android API 36 native and Blazor runtime diagnostics: 4/4 passed
- WPF sample, screenshot, diagnostics, and Inspector state verified live
- Agent, exhaustive, and CI CLI profiles verified against the sample
- DevFlow, CLI, platform collector, and sample builds complete with no warnings
- Analyzer and `buildTransitive` package assets verified

The iOS, Mac Catalyst, macOS, and GTK collectors compile successfully. Their runtime acceptance remains gated on macOS/Linux hosts, which is why the feature is not enabled by default yet.


jfversluis (changes requested):
The architecture and core premise are strong: screenshots show pixels, not causes, and this adds useful structural evidence with explicit incomplete coverage. However, I found seven issues that should be addressed before merge.

1. **[High] Mixed-case overlap rule requests can produce a false pass**

   `src/DevFlow/Microsoft.Maui.DevFlow.Agent.Core/LayoutDiagnostics/LayoutDiagnosticsEngine.cs:890-892` uses case-sensitive constant patterns to decide whether overlap analysis runs, while request validation and pass accounting are case-insensitive (`:556-575`). A valid casing variant such as `layout.Geometric-Overlap` skips analysis but is subsequently counted as passed. Please canonicalize rule IDs or use `OrdinalIgnoreCase` consistently and add a mixed-case regression test.

2. **[Medium] Finding ID assignment is quadratic before truncation**

   `LayoutDiagnosticsEngine.cs:1124-1182` scans every existing finding for every new ID. The 1,000-finding limit is applied only after analysis in `MauiDevFlowAgentService.LayoutDiagnostics.cs:176-180`. With up to 5,000 nodes and pass findings enabled, this can create roughly 35,000 findings and hundreds of millions of comparisons while holding the diagnostics gate. Please track assigned IDs in a `HashSet<string>` and enforce the result limit during analysis.

3. **[Medium] Effective alpha is only partially accumulated on Android, UIKit, and AppKit**

   `src/DevFlow/Microsoft.Maui.DevFlow.Agent/VisualTreeWalker.LayoutDiagnostics.cs:310-322`, `:449-460`, and `:564-576` stop multiplying ancestor alpha once the product drops below `0.99`, but that partial product is also compared against the `0.01` hit-test threshold. Compounded transparency can therefore classify effectively invisible elements as hit-testable. Please compute the full product for hit-test visibility.

4. **[Medium] UIKit uses `UIView.Opaque` as evidence of painted opacity**

   `VisualTreeWalker.LayoutDiagnostics.cs:370` marks a view opaque from the rendering hint alone. Plain transparent MAUI layout containers can have this flag set, producing platform-specific false `layout.visual-occluded` findings. Please require an actually opaque background/layer fill, consistent with the other collectors.

5. **[Medium] Element-detail requests repeat a full tree walk in Debug**

   `src/DevFlow/Microsoft.Maui.DevFlow.Agent.Core/MauiDevFlowAgentService.cs:540-553` resolves the element, then walks and source-aligns the full tree across all windows to transfer one source location whenever Debug source maps are active. This blocks the UI thread on an interactive Inspector path. Please attach source information directly or cache a source-mapped tree by revision.

6. **[Medium] The Driver's 501 fallback is unreachable**

   `src/DevFlow/Microsoft.Maui.DevFlow.Driver/AgentClient.cs:358-374` expects to inspect a 501 response, but shared response handling throws `NotSupportedByAgentException` first (`:1912-1964`). Unsupported backends bypass the tailored CLI/MCP capability messages. Please catch that exception in `AnalyzeLayoutAsync` and return `null`, matching the intended contract and sibling methods.

7. **[Medium] Inspector compares native-inclusive and native-excluded revisions**

   `src/Cli/Microsoft.Maui.Cli/DevFlow/Inspector/InspectorServer.cs:937` fetches a native-inclusive tree, while `:1095-1101` requests diagnostics without native elements. On Windows/WPF when a native dialog or detached top-level window is present, the revisions cannot match. The Inspector retries three times, discards diagnostics at `:1017-1019`, and overwrites the cache with null at `:1046-1049`. Please compare like-for-like revisions and avoid replacing valid cached diagnostics with null for an incoherent frame.

The shared analysis model, explicit coverage semantics, privacy handling, DTO wire compatibility, suppression persistence, and DOM rendering approach otherwise look solid.

morning4coffe-dev:
@jfversluis thanks for the careful review — all seven are addressed in `adde50b1`, plus an eighth we found while validating (`POST /webview/navigate` wrote the new URL only to the defensive snapshot copy, so `/webview/contexts` kept reporting the pre-navigation URL).

Regression tests were added for 1, 2, 6 and 8, and each was confirmed to fail with its fix reverted. 3, 4, 5 and 7 are covered by the existing suite plus a live pass against the Windows sample. Full run: 496 DevFlow + 816 CLI tests, 0 warnings.

Three things worth calling out:

**2 — finding ID assignment.** Switched to a `HashSet<string>` as suggested, which removes the quadratic scan. I did *not* enforce the 1,000-finding limit during analysis: findings are sorted by outcome/severity/confidence *before* truncation, so capping at discovery time would keep an arbitrary first 1,000 and drop violations in favour of observations. Happy to add it if you'd rather bound the intermediate allocation as well.

**7 — Inspector revision mismatch.** This turned out to be worse than described: with native elements present the coherent-frame retry never converges, so `/api/state` fails outright rather than just losing diagnostics. Reproduced live with a native dialog open (6 native-origin nodes) — before the fix, `{"error":"Unable to capture the current app state.","retryable":true}` on every attempt; after, 3/3 successful captures with diagnostics attached. Both halves are in: like-for-like revisions, and the cache is no longer overwritten with null for an incoherent frame.

**4 — opaque hints.** Applied to AppKit as well as UIKit. `NSView.IsOpaque` defaults to false so it failed safe rather than producing false findings, but it was the same "rendering hint treated as painted fill" pattern — both now read the actual background/layer alpha, consistent with the Android and WinUI collectors.

One caveat: 3 and 4 are compile-verified across all Apple TFMs but not runtime-verified on device in this pass. WinUI was deliberately left untouched, since its early break only feeds a `>= 0.99` test.

jfversluis (approved):
Re-reviewed the follow-up commit. All seven previously reported issues are resolved, and the additional WebView URL-state fix is correct. The new regression tests meaningfully cover the mixed-case rule handling, finding-ID uniqueness path, unsupported-capability fallback, and WebView registry update.

I also verified the updated CI matrix is fully green. Approving.

PR #432, opened 2026-08-15
fix(devflow): make background-job capability reporting truthful; add real WorkManager/BGTask sample jobs

aritchie:
## Why

DevFlow's background-job API (`/api/v1/device/jobs`, `maui_jobs_list`, `maui_jobs_run`) was wired end to end but had never actually returned a job. Nothing in the repo exercised it: `DevFlow.Sample` registers no WorkManager Worker and no BGTask, so `Jobs_ReturnsSupportedFlagAndJobArray` asserted only *shape* — `supported` is a bool, `jobs` is an array. An empty array from a total failure passed identically to a working query.

Putting real jobs in the sample turned up three genuine bugs and one platform limit that were invisible until something real was on the other end. All three are fixed here.

### 1. Android reported `supported: true` when WorkManager was absent

`IsJobsSupported` was hardcoded `true` for Android. WorkManager is an opt-in AndroidX dependency, not part of the platform, so an app that never referenced it was told the capability existed. Worse, the failure path *also* returned `supported: true` with an `error`, so a missing dependency was indistinguishable from a healthy, idle queue.

Now gated on an actual classpath probe, and three states are distinguishable:

| State | Response |
|---|---|
| WorkManager absent | `supported: false` + `reason` naming the package to add |
| Present, not initialized | `supported: true` + `error: "WorkManager not initialized"` |
| Present and queried | `supported: true` + jobs |

"Present but uninitialized" stays `supported: true` deliberately — that's a config problem in an app that *does* have the capability, and separating it from "dependency missing" tells you which thing to go fix.

### 2. `Java.Lang.Class.ForName(string)` could never find WorkManager — in any app

This is the more serious one, and it only surfaced once the sample genuinely had WorkManager: the agent *still* reported "not on the classpath".

The single-argument `Class.ForName` overload resolves against the **calling class's** loader. When the call originates from mono over JNI, that is the boot class loader, which cannot see anything the app or its AndroidX dependencies contribute. It throws `ClassNotFoundException` for `androidx.work.WorkManager` even in an app plainly using WorkManager.

All four call sites now resolve through `Application.Context.ClassLoader` via a `FindAndroidClass` helper. Verified on an emulator: `supported` flips to `true` once the app really has WorkManager.

### 3. Nothing to test against

- **Android** — `SampleSyncWorker`, a real `Worker` enqueued on launch with a long initial delay so it sits pending in JobScheduler rather than firing on its own. It records run count / result / timestamp in `Preferences` so a test can assert the job *executed*, not merely that it was triggered. Supports an input flag to force the failure path.
- **iOS** — `SampleBackgroundTask`, a real `BGProcessingTaskRequest` with a registered handler, `BGTaskSchedulerPermittedIdentifiers` and `UIBackgroundModes` added to `Info.plist`, and registration wired into `AppDelegate.FinishedLaunching` (it must happen before that returns). Exposed through `[DevFlowAction]`s so the whole flow is drivable over the existing invoke API.

### 4. Tests that can fail

The old test asserted only that `supported` was a bool and `jobs` was an array, which a totally broken query satisfies. Now:

- `Jobs_Android_ListsTheSampleWorker` asserts the sample worker actually comes back — by tag, with a non-empty UUID and a valid state. This is the assertion that exercises the reflection path; shape-only checks passed indefinitely while the query silently returned nothing.
- Any `error` in the payload now fails the test outright, so a degraded query can't pass as an idle queue.
- `Jobs_UnsupportedCapability_IsReportedConsistently` cross-checks the jobs list against the capabilities document, so a caller feature-detecting one way can't get a different answer than one reading the other.

### 5. Android job listing returned `[]` even with work scheduled

The third bug, and the one that made the feature useless. `Java.Lang.Reflect.Method.Invoke` hands back a plain `Java.Lang.Object` wrapper regardless of the real runtime type, so `as Java.Util.IList` and `is Java.Util.ICollection` **always** failed. The null checks then swallowed the result — an unreachable list was indistinguishable from an idle queue.

- Reflected results are re-wrapped by JNI handle (`AsJavaInterface`) so the work-info list and tag set bind to their real types.
- Boxed primitives are read via their string form; they are opaque wrappers too.
- `java.util.List` is resolved by name — `Class.FromType(typeof(Java.Util.IList))` does not reliably map to it.
- `ListenableFuture.get()` is bounded with a 5s timeout in C# rather than the `get(long, TimeUnit)` overload, whose primitive `long` parameter cannot be resolved by `Class.FromType`.
- **Every failure path now returns an explicit `error`.** An empty `jobs` array now means exactly one thing: no work is scheduled.

## Verified on real hardware

**Android emulator (API 37)** — forcing a WorkManager job over adb works, end to end:

```
GET /api/v1/device/jobs
  { "supported": true, "jobs": [ { "identifier": "c7c3961d-4c6d-49eb-aee1-2266229871ee",
                                   "tags": ["devflow-sample-sync", "…SampleSyncWorker"],
                                   "state": "ENQUEUED", "runAttemptCount": 0 } ] }

$ adb shell cmd jobscheduler run -f -n androidx.work.systemjobscheduler com.companyname.mauitodo 3
Running job [FORCED]

logcat: WM-WorkerWrapper: Worker result SUCCESS for Work [ id=…, tags={…, devflow-sample-sync} ]

GET /api/v1/device/jobs
  { … "state": "SUCCEEDED", "runAttemptCount": 1 }
```

The full loop — list, trigger, observe the outcome — now works on Android.

Notes for whoever builds on this: the `-n androidx.work.systemjobscheduler` namespace is **required** (WorkManager schedules into it; without `-n` the job isn't found), job ids are small per-app integers that must be read from `dumpsys jobscheduler` rather than assumed, and `cmd jobscheduler` also offers `get-job-state`, `stop` and `timeout` — so the `onStopped` and expiration paths are reachable too.

**iPhone 15 Pro / iOS 26.6** — the full BGTask chain:

```
[DevFlow.Sample] BGTask schedule: submitted

GET /api/v1/device/jobs
  { "supported": true, "runSupported": true,
    "jobs": [ { "identifier": "com.companyname.mauitodo.sync",
                "type": "processing",
                "earliestBeginDate": "2026-08-16 20:53:51 +0000" } ] }

POST /invoke/actions/bgtask-simulate-launch
[DevFlow.Sample] BGTask 'com.companyname.mauitodo.sync' executed.
  run count 0 → 1 · pending jobs 1 → 0   (request consumed, exactly as a real launch)
```

This is the first time the iOS listing path has returned a real job. Reaching the agent over USB needs `iproxy 19226:19226` — `devicectl` has no port forwarding, which is worth knowing for any physical-device DevFlow work, not just jobs.

**iOS Simulator — BGTasks do not work at all.** `BGTaskScheduler.Submit` fails with `BGTaskSchedulerErrorCodeUnavailable (1)`. Handlers register fine; nothing after that works, `GetPending` is permanently empty, and there is nothing for a simulated launch to find. **iOS background-job testing requires a physical device**; it cannot run on simulator CI. That is a platform constraint, not something DevFlow can engineer around.

*(If you ever see `NotPermitted (3)` instead: the incremental iOS build silently fails to re-copy `Info.plist` into the app bundle. Check the bundle's plist, not the source.)*

Also run: 226 DevFlow unit tests, 463 CLI unit tests, and the `DeviceTests` integration suite on both an Android emulator (13/13) and an iOS Simulator (13/13). Android and iOS sample builds clean.

## CLI: list and run

`maui devflow ui jobs list` and `maui devflow ui jobs run <id>`. Listing worked over the agent but had no CLI surface, so it was effectively unusable. Running reports the real outcome, and `run` returns its own exit code so it is usable as a CI assertion.

```
$ maui devflow ui jobs list
Android / WorkManager: 1 job(s)
  3025764d-c312-478c-8847-b154e8e7ae5a
    state:    ENQUEUED
    attempts: 0
    tags:     devflow-sample-sync, crc6448dcefd8c32368e1.SampleSyncWorker

$ maui devflow ui jobs run 3025764d-c312-478c-8847-b154e8e7ae5a
Forced job 9 (SampleSyncWorker) via adb — Running job [FORCED]
SUCCESS: SUCCEEDED after 1 attempt(s)                                    # exit 0

$ maui devflow ui jobs list          # on iPhone 15 Pro
iOS / BGTaskScheduler: 1 job(s)
  com.companyname.mauitodo.sync
    type:     processing
    earliest: 2026-08-16 21:50:38 +0000

$ maui devflow ui jobs run com.companyname.mauitodo.sync
SUCCESS: BGTask 'com.companyname.mauitodo.sync' was launched and its pending request was consumed.
```

**Android** — a job cannot be forced from inside the app, because the original `WorkRequest` cannot be reconstructed from a `WorkInfo`. adb can force it through JobScheduler, so `jobs run` triggers host-side and then polls the in-app WorkManager query for the terminal state — the only place the real SUCCEEDED/FAILED result lives. `AndroidAppDriver` gains `ListScheduledJobsAsync`, `GetJobStateAsync`, `RunScheduledJobAsync` and `StopScheduledJobAsync`; the `dumpsys` parser handles both the namespaced and plain job forms and is unit tested against captured emulator output. A WorkManager UUID is mapped to its numeric job id via the worker name WorkManager auto-tags onto every request.

**iOS** — `RunPlatformJobAsync` schedules the request only if it is not already pending, then forces the launch with `_simulateLaunchForTaskWithIdentifier:`. Submitting alone only *schedules*; iOS may never launch it. Success is judged by the pending request being consumed, which is the observable evidence a launch really happened. **The selector is compiled out of Release builds** — the agent ships inside consumer apps and App Store review scans for private selectors, so release builds can schedule but not trigger. The advertised capability is gated on the same condition: `runSupported` and the `run` feature both read `false` in a Release agent, so a caller feature-detecting the operation never gets an answer the binary cannot honour.

All four outcome paths verified on hardware:

| | result | exit |
|---|---|---|
| Android, succeeding worker | `SUCCESS: SUCCEEDED after 1 attempt(s)` | 0 |
| Android, failing worker | `FAILED: FAILED after 1 attempt(s)` | 1 |
| iOS, registered identifier | `SUCCESS: … pending request was consumed` | 0 |
| iOS, unregistered identifier | `FAILED: … not permitted. Add it to BGTaskSchedulerPermittedIdentifiers …` | 1 |

Two further fixes came out of that verification. Every BGTask submit failure had been reported as "unavailable on the iOS Simulator", true only for error code 1 — a real device returning code 3 (NotPermitted) was being blamed on the simulator. The three codes now map to their actual causes. And `jobs run` exited 0 on failure, because `_errorOccurred` is only consumed by the batch-script path; `jobs run` now returns its own exit code. That pre-existing behaviour is left alone for every other command rather than changing CLI-wide exit semantics here.

## Known gaps — deliberately not fixed here

- **The agent's `runSupported` flag still reads `false` on Android**, which is accurate for the in-app API but understates what the CLI can do via adb. Worth reconciling into a richer capability shape.
- **`maui devflow ui jobs list` needs an explicit `--agent-port`.** The option defaults to 9223 while the sample's agent starts on 9225, and the mismatch surfaces as `{"error":"Operation is not valid due to the current state of the object."}`, which says nothing about the real cause. Pre-existing and not specific to jobs.

## Dependency note

`Xamarin.AndroidX.Work.Runtime` is pinned to **2.10.0.5**, the newest version whose *complete* dependency graph is available on this repo's feeds.

`NuGet.config` clears all sources and restores from the dnceng feeds alone. From WorkManager 2.10.1 on, the graph pulls Room 2.7+, which AndroidX split into KMP-style sub-packages — `Room.Runtime.Android`, `Room.Common.JVM`, `Sqlite.Android`, `Sqlite.Framework.Android`. Those are **not mirrored to dotnet-public**, so a cold restore fails with NU1101 even though the same restore succeeds on a machine with a warm package cache. Room 2.6.x, which 2.10.0.5 uses, predates the split.

Two follow-on problems came with that pin, both settled by referencing the matching 1.3.0 packages the MAUI baseline already resolves:

| Problem | Cause | Fix |
|---|---|---|
| NU1107 | 2.10.0.5 wants `Concurrent.Futures.Ktx` 1.2.0.7, constraining the base to `< 1.2.1`, while its own `Core` dependency wants the base `>= 1.3.0.1` | reference `Concurrent.Futures.Ktx` 1.3.0.1 |
| D8 dexing | *"androidx.tracing.TraceKt$traceAsync$1 is defined multiple times"* — the same ktx-merged-into-base collision seen with Core 1.19 | reference `Tracing.Tracing.Ktx` 1.3.0.1 |

`Lifecycle.Service` is matched to the baseline `Lifecycle.Runtime` for the same reason, which clears the last NU1608. Rationale is documented inline in `eng/Versions.props`; revisit once the Room/Sqlite platform packages are mirrored, which would allow moving back to the 2.11 line.

Verified with an **empty package cache against the approved feeds only** — the CI condition — restore and Android build succeed with no errors and no NuGet warnings.

## Re-verified after the version change

Downgrading WorkManager moves Room from 2.8.x to 2.6.x, and Room is where WorkManager stores job state and input data, so the whole loop was re-run on an emulator (Pixel 9a, API 37):

```
GET /api/v1/device/jobs
  { "supported": true, "runSupported": false,
    "jobs": [ { "identifier": "b232656a-25aa-4fb2-a316-e955f8f65b59",
                "tags": ["devflow-sample-sync", "…SampleSyncWorker"],
                "state": "ENQUEUED", "runAttemptCount": 0 } ] }

$ maui devflow ui jobs run b232656a-… --agent-port 9225
Forced job 0 (SampleSyncWorker) via adb — Running job [FORCED]
SUCCESS: SUCCEEDED after 1 attempt(s)                                    # exit 0
```

`/api/v1/agent/capabilities` agrees with the listing (`supported: true`, `features: ["list"]`), and all four `jobs run` outcome paths still return the exit codes in the table above.

The part that actually mattered is what the worker recorded for itself, since that is the evidence the job *executed* rather than merely being triggered:

```
worker-last-result   success  →  failure     (after a shouldFail=true run)
worker-run-count     11       →  12
worker-last-run-utc  2026-08-31T12:16:26.5953422Z
```

The `shouldFail` flag round-tripped through Room storage into `DoWork`, the body ran, and the outcome came back out through the reflected query — so the older Room line carries the feature intact.

### DIFF
--- a/samples/DevFlow.Sample/DevFlow.Sample.csproj
+++ b/samples/DevFlow.Sample/DevFlow.Sample.csproj
@@ -38,6 +38,14 @@
     <!-- WorkManager gives DevFlow a real background job to list, force-run via adb, and assert on. -->
     <ItemGroup Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">
         <PackageReference Include="Xamarin.AndroidX.Work.Runtime" />
+        <!-- Work.Runtime asks for 1.2.x of these two -Ktx packages, whose classes AndroidX has
+             since merged into their base packages. The MAUI baseline pulls the merged 1.3.0
+             bases in, so the old -Ktx packages both conflict on version and duplicate classes
+             at dex time. Referencing the matching 1.3.0 -Ktx facades directly settles both.
+             Lifecycle.Service is matched to the baseline Lifecycle.Runtime for the same reason. -->
+        <PackageReference Include="Xamarin.AndroidX.Concurrent.Futures.Ktx" />
+        <PackageReference Include="Xamarin.AndroidX.Tracing.Tracing.Ktx" />
+        <PackageReference Include="Xamarin.AndroidX.Lifecycle.Service" />
     </ItemGroup>
 
     <ItemGroup>
--- a/samples/DevFlow.Sample/DevFlow.Sample.csproj
+++ b/samples/DevFlow.Sample/DevFlow.Sample.csproj
@@ -35,6 +35,11 @@
         <PackageReference Include="Microsoft.Extensions.Logging.Debug" />
     </ItemGroup>
 
+    <!-- WorkManager gives DevFlow a real background job to list, force-run via adb, and assert on. -->
+    <ItemGroup Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">
+        <PackageReference Include="Xamarin.AndroidX.Work.Runtime" />
+    </ItemGroup>
+
     <ItemGroup>
         <ProjectReference Include="..\..\src\DevFlow\Microsoft.Maui.DevFlow.Agent\Microsoft.Maui.DevFlow.Agent.csproj" />
         <ProjectReference Include="..\..\src\DevFlow\Microsoft.Maui.DevFlow.Blazor\Microsoft.Maui.DevFlow.Blazor.csproj" />
--- a/src/DevFlow/Microsoft.Maui.DevFlow.Agent/DevFlowAgentService.cs
+++ b/src/DevFlow/Microsoft.Maui.DevFlow.Agent/DevFlowAgentService.cs
@@ -583,23 +583,66 @@ public class PlatformAgentService : DevFlowAgentService
         try
         {
             var taskType = await ResolveBgTaskRequestTypeAsync(identifier, type);
-            BGTaskRequest taskRequest = taskType.Equals("refresh", StringComparison.OrdinalIgnoreCase)
-                ? new BGAppRefreshTaskRequest(identifier)
-                : new BGProcessingTaskRequest(identifier);
 
-            taskRequest.EarliestBeginDate = null;
-
-            BGTaskScheduler.Shared.Submit(taskRequest, out var error);
-            if (error != null)
-                return new { success = false, error = error.LocalizedDescription, identifier };
-
-            return await Task.FromResult<object?>(new
+            // The request must be pending before it can be launched: iOS launches a *submitted*
+            // task, and simulating a launch for an unscheduled identifier is a no-op.
+            var wasPending = await IsBgTaskPendingAsync(identifier);
+            string? submitError = null;
+            if (!wasPending)
             {
-                success = true,
-                message = $"BGTask '{identifier}' submitted",
+                BGTaskRequest taskRequest = taskType.Equals("refresh", StringComparison.OrdinalIgnoreCase)
+                    ? new BGAppRefreshTaskRequest(identifier)
+                    : new BGProcessingTaskRequest(identifier);
+                taskRequest.EarliestBeginDate = null;
+
+                BGTaskScheduler.Shared.Submit(taskRequest, out var error);
+                submitError = error?.LocalizedDescription;
+            }
+
+            if (submitError != null)
+            {
+                return new
+                {
+                    success = false,
+                    identifier,
+                    type = taskType,
+                    error = $"Could not schedule BGTask '{identifier}': {submitError}. " +
+                            "BGTaskScheduler is unavailable on the iOS Simulator — background jobs require a physical device."
+                };
+            }
+
+            // Submitting only *schedules*; iOS decides when to launch, which organically may be
+            // never. Forcing the launch is the only way to actually run the handler on demand.
+            var launched = TrySimulateBgTaskLaunch(identifier, out var launchError);
+            if (!launched)
+            {
+                return new
+                {
+                    success = false,
+                    identifier,
+                    type = taskType,
+                    scheduled = true,
+                    error = launchError
+                };
+            }
+
+            // The scheduler consumes the pending request when it launches it, so the request
+            // disappearing is the observable evidence that a launch really happened.
+            await Task.Delay(1500);
+            var stillPending = await IsBgTaskPendingAsync(identifier);
+
+            return new
+            {
+                success = !stillPending,
                 identifier,
-                type = taskType
-            });
+                type = taskType,
+                launched = true,
+                consumedPendingRequest = !stillPending,
+                message = stillPending
+                    ? $"BGTask '{identifier}' launch was requested but the request is still pending; the handler may not be registered."
+                    : $"BGTask '{identifier}' was launched and its pending request was consumed.",
+                note = "iOS does not expose a task's result; assert on what the handler itself records."
+            };
         }
         catch (Exception ex)
         {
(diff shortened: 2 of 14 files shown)


samples/DevFlow.Sample/DevFlow.Sample.csproj
```diff
@@ -35,6 +35,11 @@
         <PackageReference Include="Microsoft.Extensions.Logging.Debug" />
     </ItemGroup>
 
+    <!-- WorkManager gives DevFlow a real background job to list, force-run via adb, and assert on. -->
+    <ItemGroup Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">
+        <PackageReference Include="Xamarin.AndroidX.Work.Runtime" />
```
jfversluis, line 40: Adding this Work.Runtime reference currently makes the sample restore fail with the repository's configured feeds. `dotnet build samples/DevFlow.Sample/DevFlow.Sample.csproj -f net10.0-android` reports NU1101 for transitive `Xamarin.AndroidX.Room.Runtime.Android`, `Xamarin.AndroidX.Room.Common.JVM`, `Xamarin.AndroidX.Sqlite.Framework.Android`, and `Xamarin.AndroidX.Sqlite.Android`. Please use a WorkManager package/version whose complete dependency graph is available to the approved feeds (or otherwise fix the dependency versions) so the sample remains buildable in CI.

src/DevFlow/Microsoft.Maui.DevFlow.Agent/DevFlowAgentService.cs
```diff
@@ -470,6 +654,84 @@ protected override bool IsJobRunSupported
...
+            error = null;
+            return true;
+        }
+        catch (Exception ex)
+        {
+            error = $"Failed to launch BGTask: {ex.GetBaseException().Message}";
+            return false;
+        }
+#else
+        error = "Triggering a BGTask requires a private API and is only available in Debug builds of the DevFlow agent. " +
```
jfversluis, line 744: This release-build branch guarantees that `RunPlatformJobAsync` returns `success=false`, but the capability response still advertises `runSupported=true` and includes the `run` feature for iOS/Mac Catalyst. Consumers will feature-detect a supported operation that is guaranteed to fail in the normal Release build; gate this capability on the same build condition (or provide a real release-safe trigger).

jfversluis (changes requested):
Requesting changes for the two correctness issues noted inline.

src/DevFlow/Microsoft.Maui.DevFlow.Agent/DevFlowAgentService.cs
aritchie: Fixed

PR #434, opened 2026-08-15
fix(devflow): invoke API crashes on Android when any DevFlowAction exists

### LINKED ISSUE #433: DevFlow: invoke API returns ArgumentException on Android whenever any [DevFlowAction] exists
## Summary

On Android, the entire invoke API fails as soon as the app declares **any** `[DevFlowAction]`:

```
GET  /api/v1/invoke/actions          → empty response
POST /api/v1/invoke/actions/{name}   → empty response

logcat: [Microsoft.Maui.DevFlow.Agent] Request error:
        ArgumentException: The value cannot be an empty string. (Parameter 'path')
```

iOS, Mac Catalyst and desktop are unaffected.

## Why it has gone unnoticed

There is currently **no `[DevFlowAction]` anywhere in `samples/` or `playground/`**, and the crash only happens when at least one action exists. With zero actions the ranking code below is never reached, so every existing test and manual check passes. The first action added to the sample broke the endpoint immediately.

## Root cause

`DevFlowAgentService.Invoke.cs` → `IsAssemblyInAppBaseDirectory`:

```csharp
private static bool IsAssemblyInAppBaseDirectory(Assembly assembly)
{
    var location = assembly.Location;
    if (string.IsNullOrEmpty(location))
        return false;

    return string.Equals(
        Path.GetFullPath(Path.GetDirectoryName(location) ?? string.Empty).TrimEnd(...),
        Path.GetFullPath(AppContext.BaseDirectory).TrimEnd(...),
        StringComparison.OrdinalIgnoreCase);
}
```

`Path.GetDirectoryName` returns an **empty string, not null**, when the path has no directory component. The `?? string.Empty` therefore never fires, and `Path.GetFullPath("")` throws `ArgumentException: The value cannot be an empty string. (Parameter 'path')`.

On Android, assemblies loaded from the APK report a bare file name with no directory, so this path is hit. `AppContext.BaseDirectory` is likewise not guaranteed to be populated on every host, and would throw the same way.

The blast radius is larger than the helper suggests. It is reached from `SelectPreferredAction`, which `ScanActions` applies to **every** action group, not just duplicates:

```csharp
return actions
    .GroupBy(a => a.Name, StringComparer.OrdinalIgnoreCase)
    .Select(SelectPreferredAction)     // ← runs for every action, even a single one
    .ToArray();
```

So one unrankable assembly takes down action discovery entirely — and because `s_cachedActions` is a `Lazy<T>` created with `ExecutionAndPublication`, the faulted result is cached and every subsequent request rethrows.

This is only a **tie-breaker for ordering duplicate action names**. It should never be able to fail the request.

## Reproduction

1. Add any `[DevFlowAction]` to a MAUI app referencing `Microsoft.Maui.DevFlow.Agent`:
   ```csharp
   [DevFlowAction("hello", Description = "test")]
   public static string Hello() => "hi";
   ```
2. Deploy to an Android emulator or device.
3. `curl http://localhost:<agentPort>/api/v1/invoke/actions`

**Expected:** the action list.
**Actual:** empty response; `ArgumentException` in logcat.

Verified on an Android emulator (API 37, Pixel 9a) with `Microsoft.Maui.DevFlow.Agent` built from `main`.

## Impact

`maui_invoke_action` / `maui_list_actions` and the `maui devflow invoke` surface are unusable on Android. DevFlow Actions are the documented way to expose app-side test hooks to AI agents, so this removes that capability on one of the two main platforms.

## Fix

Guard the empty-directory case, guard an empty `AppContext.BaseDirectory`, and make the helper non-throwing — it is a ranking hint, not a correctness requirement.

aritchie:
Fixes #433

## The bug

On Android, the whole invoke API fails as soon as the app declares **any** `[DevFlowAction]`:

```
GET  /api/v1/invoke/actions          → empty response
POST /api/v1/invoke/actions/{name}   → empty response

logcat: [Microsoft.Maui.DevFlow.Agent] Request error:
        ArgumentException: The value cannot be an empty string. (Parameter 'path')
```

`Path.GetDirectoryName` returns an **empty string, not null**, when a path has no directory component, so the `?? string.Empty` never fired and `Path.GetFullPath("")` threw. Android assemblies loaded from the APK report a bare file name with no directory, so this was hit on every request. `AppContext.BaseDirectory` is likewise not guaranteed to be populated and would fail identically.

The blast radius is wider than the helper looks. `SelectPreferredAction` is applied by `ScanActions` to **every** action group, not just duplicated names:

```csharp
return actions
    .GroupBy(a => a.Name, StringComparer.OrdinalIgnoreCase)
    .Select(SelectPreferredAction)     // ← every action, even a single one
    .ToArray();
```

So one unrankable assembly failed discovery entirely — and since `s_cachedActions` is a `Lazy<T>` created with `ExecutionAndPublication`, the faulted result was cached and every subsequent request rethrew it.

## The fix

This value is only a **tie-breaker for ordering duplicate action names**. It should never be able to fail a request, so it now guards both empty inputs and swallows malformed-path exceptions.

The path comparison is extracted into `internal static bool IsInAppBaseDirectory(string?, string?)` so the logic is testable without a device — the previous form needed a real `Assembly` with a pathological `Location`, which is why it had no coverage.

## Why this was never caught

There is **no `[DevFlowAction]` anywhere in `samples/` or `playground/`** on `main`, and the crash only fires when at least one action exists. With zero actions the ranking code is never reached, so everything passed. The first action added to the sample broke the endpoint immediately.

## Tests

9 new tests in `InvokeActionRankingTests`, covering the Android bare-filename case, empty and null locations, empty and null base directories, malformed paths, and the positive match and trailing-separator cases.

I verified they actually catch the bug rather than just passing: reverting the helper to the previous implementation fails **4 of the 9**; the fix takes it back to 9/9. Full suite 233/233.

## Device verification

Android emulator, API 37 (Pixel 9a), with a `[DevFlowAction]` present:

```
$ curl .../api/v1/invoke/actions
{ "actions": [ { "name": "workmanager-enqueue", "declaringType": "DevFlow.Sample.SampleSyncWorker",
                 "parameters": [ { "name": "shouldFail", "type": "bool", … } ] } ] }

$ curl -X POST .../api/v1/invoke/actions/workmanager-enqueue -d '{"args":[true]}'
{ "success": true, "returnValue": "enqueued 'devflow-sample-sync' (shouldFail=True)" }

logcat "empty string" errors: 0
```

## Note on scope

The sample app still declares no `[DevFlowAction]`, so this endpoint remains uncovered by the integration suite on `main`. #432 adds the first ones (`workmanager-enqueue`, and the iOS BGTask actions) — that is the branch I used to reproduce and verify this fix, since a bare `main` cannot trigger the bug at all. Worth adding a sample action plus an integration assertion regardless of which PR lands first, so this cannot regress silently again.

### DIFF
--- a/src/DevFlow/Microsoft.Maui.DevFlow.Agent.Abstractions/DevFlowActionAttribute.cs
+++ b/src/DevFlow/Microsoft.Maui.DevFlow.Agent.Abstractions/DevFlowActionAttribute.cs
@@ -39,7 +39,8 @@ namespace Microsoft.Maui.DevFlow.Agent.Core;
 /// </list>
 ///
 /// <para>
-/// Methods must be <c>public static</c>. Return type should be <c>void</c>,
+/// Methods must be <c>public static</c> and must not contain generic parameters; this excludes
+/// generic methods and methods declared on generic types. Return type should be <c>void</c>,
 /// <c>Task</c>, or <c>Task&lt;T&gt;</c> where T is a supported type or any
 /// type whose <c>ToString()</c> produces a meaningful result.
 /// </para>
--- a/src/DevFlow/Microsoft.Maui.DevFlow.Agent.Abstractions/DevFlowAgentService.Invoke.cs
+++ b/src/DevFlow/Microsoft.Maui.DevFlow.Agent.Abstractions/DevFlowAgentService.Invoke.cs
@@ -69,8 +69,12 @@ public partial class DevFlowAgentService
 			{
 				foreach (var method in type.GetMethods(BindingFlags.Public | BindingFlags.Static))
 				{
+					if (method.ContainsGenericParameters)
+						continue;
+
 					var attr = method.GetCustomAttribute<DevFlowActionAttribute>();
-					if (attr == null) continue;
+					if (attr == null)
+						continue;
 
 					actions.Add(new InvokeActionEntry
 					{
@@ -137,19 +141,19 @@ public partial class DevFlowAgentService
 		if (string.IsNullOrEmpty(assemblyLocation) || string.IsNullOrEmpty(appBaseDirectory))
 			return false;
 
-		// Empty, not null, when the location carries no directory component.
-		var directory = Path.GetDirectoryName(assemblyLocation);
-		if (string.IsNullOrEmpty(directory))
-			return false;
-
 		try
 		{
+			// Empty, not null, when the location carries no directory component.
+			var directory = Path.GetDirectoryName(assemblyLocation);
+			if (string.IsNullOrEmpty(directory))
+				return false;
+
 			return string.Equals(
 				Path.GetFullPath(directory).TrimEnd(Path.DirectorySeparatorChar),
 				Path.GetFullPath(appBaseDirectory).TrimEnd(Path.DirectorySeparatorChar),
 				StringComparison.OrdinalIgnoreCase);
 		}
-		catch (Exception ex) when (ex is ArgumentException or NotSupportedException or PathTooLongException)
+		catch (Exception ex) when (ex is ArgumentException or NotSupportedException or PathTooLongException or System.Security.SecurityException)
 		{
 			// Malformed or non-filesystem paths are not worth failing action discovery over.
 			return false;
--- a/src/DevFlow/Microsoft.Maui.DevFlow.Agent.Abstractions/DevFlowAgentService.Invoke.cs
+++ b/src/DevFlow/Microsoft.Maui.DevFlow.Agent.Abstractions/DevFlowAgentService.Invoke.cs
@@ -116,15 +116,44 @@ public partial class DevFlowAgentService
 			.First();
 
 	private static bool IsAssemblyInAppBaseDirectory(Assembly assembly)
+		=> IsInAppBaseDirectory(assembly.Location, AppContext.BaseDirectory);
+
+	/// <summary>
+	/// Whether an assembly location sits in the app base directory. This is only a tie-breaker
+	/// for ranking duplicate action names, so anything it cannot determine is simply "no".
+	///
+	/// <para>
+	/// It must never throw. Every DevFlow action is ranked through here, so an exception takes
+	/// down <c>GET /api/v1/invoke/actions</c> and every <c>POST …/actions/{name}</c> with it.
+	/// On Android, assemblies loaded from the APK report a bare file name with no directory
+	/// component, and <see cref="Path.GetDirectoryName(string)"/> returns an empty string rather
+	/// than null for those — which the previous <c>?? string.Empty</c> could not catch, so
+	/// <see cref="Path.GetFullPath(string)"/> threw
+	/// <c>ArgumentException: The value cannot be an empty string. (Parameter 'path')</c>.
+	/// </para>
+	/// </summary>
+	internal static bool IsInAppBaseDirectory(string? assemblyLocation, string? appBaseDirectory)
 	{
-		var location = assembly.Location;
-		if (string.IsNullOrEmpty(location))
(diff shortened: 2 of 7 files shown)


src/DevFlow/Microsoft.Maui.DevFlow.Agent.Core/DevFlowAgentService.Invoke.cs
```diff
@@ -116,15 +116,44 @@ private static InvokeActionEntry SelectPreferredAction(IEnumerable<InvokeActionE
...
-		if (string.IsNullOrEmpty(location))
+		if (string.IsNullOrEmpty(assemblyLocation) || string.IsNullOrEmpty(appBaseDirectory))
 			return false;
 
-		return string.Equals(
-			Path.GetFullPath(Path.GetDirectoryName(location) ?? string.Empty).TrimEnd(Path.DirectorySeparatorChar),
-			Path.GetFullPath(AppContext.BaseDirectory).TrimEnd(Path.DirectorySeparatorChar),
-			StringComparison.OrdinalIgnoreCase);
+		// Empty, not null, when the location carries no directory component.
+		var directory = Path.GetDirectoryName(assemblyLocation);
```
jfversluis: `Path.GetDirectoryName(assemblyLocation)` is still outside the `try`, so the helper does not actually uphold the documented “must never throw” contract for malformed paths. On Windows, `Path.GetDirectoryName` can throw `ArgumentException` for invalid path characters (for example the NUL-containing path covered by `IsInAppBaseDirectory_ToleratesMalformedPaths`), causing action discovery to fail before the catch around `GetFullPath` is reached. Move this call into the guarded block (or catch its exceptions too) so every path operation is protected.

jfversluis (changes requested):
Requesting changes: malformed path handling in IsInAppBaseDirectory can still throw before the try/catch and crash action discovery on invalid Windows paths.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Core/DevFlowAgentService.Invoke.cs
```diff
@@ -116,15 +116,44 @@ private static InvokeActionEntry SelectPreferredAction(IEnumerable<InvokeActionE
...
+			return string.Equals(
+				Path.GetFullPath(directory).TrimEnd(Path.DirectorySeparatorChar),
+				Path.GetFullPath(appBaseDirectory).TrimEnd(Path.DirectorySeparatorChar),
+				StringComparison.OrdinalIgnoreCase);
+		}
+		catch (Exception ex) when (ex is ArgumentException or NotSupportedException or PathTooLongException)
+		{
+			// Malformed or non-filesystem paths are not worth failing action discovery over.
+			return false;
+		}
```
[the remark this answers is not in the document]
Redth: Fixed in c832d9d8. `Path.GetDirectoryName` now runs inside the guarded block, and `SecurityException` is handled alongside the other path exceptions. The Windows CI job passes with the malformed-path coverage.

src/DevFlow/Microsoft.Maui.DevFlow.Agent.Core/DevFlowAgentService.Invoke.cs
Redth: Fixed in c832d9d8 by moving `Path.GetDirectoryName` into the guarded block. The existing NUL-path regression test now covers the full operation on Windows, and the Windows CI job passes.

jfversluis (approved):
Approved after re-review: the malformed-path fix is in place, the action-discovery hardening looks correct, and I did not find any remaining high-confidence correctness or security issues. Validation in this environment was limited by a local workload-install/storage issue, but the updated logic and regression coverage are consistent with the fix.

Issue #341, opened 2026-06-22
macOS AppKit: agent connection refused on localhost when agent binds IPv4 and CLI resolves ::1


davidortinau:
## Summary

On macOS **AppKit**, the agent binds IPv4 `127.0.0.1` but the CLI's default `localhost` resolves to IPv6 `::1` first, so `ui`/`webview` commands report "connection refused" even though the agent is listening. Same IPv4/IPv6 root cause as #268 (observed there for Android emulator + `adb forward`); this is the macOS direct-localhost variant.

## Environment

- Host OS/platform: macOS 26.5.1 (arm64)
- AI tool/host: GitHub Copilot CLI
- Model: Claude Opus 4.8
- MAUI CLI: `maui devflow` 0.1.0-preview.11.26317.2 (CLI commands, not MCP)
- App framework(s): .NET MAUI 11.0.0, Blazor Hybrid (WebView)
- Target platform(s): macOS AppKit (`net11.0-macos`) via community `Platform.Maui.MacOS` 0.3.0 + `Platform.Maui.MacOS.BlazorWebView` 0.3.0
- .NET SDK: 11.0.100-preview.5
- MAUI DevFlow package versions: `Microsoft.Maui.DevFlow.Agent` 0.25.0-dev, `Microsoft.Maui.DevFlow.Blazor` 0.25.0-dev; agent runtime 0.25.0-dev

## What happened

After an app relaunch, `ui status` and all `webview` commands returned "Cannot connect to agent at localhost:&lt;port&gt;" / "Connection refused (localhost:&lt;port&gt;)", which looked like a dead agent.

## Attempts and workarounds

1. Repeated `wait` / `agent status` / `webview status` — all refused on `localhost`.
2. Inspected the live socket: the agent was LISTENING on IPv4 `127.0.0.1:&lt;port&gt;`.
3. Passing `--agent-host 127.0.0.1` (forcing IPv4) connected immediately for both the agent and CDP.

## Expected behavior

For `localhost`, the CLI should attempt both `127.0.0.1` and `::1` (happy-eyeballs), or the agent should bind both address families, so the default works without a manual host override.

## Actual behavior

Default `localhost` resolution hit IPv6 `::1` and was refused while the agent listened on IPv4; the failure was indistinguishable from a crashed/missing agent until the socket was inspected.

## Outcome

Worked around with `--agent-host 127.0.0.1`. This was the single biggest time sink in the session.

## Scope

[...]

PR #438, opened 2026-08-18
Split DevFlow Driver into portable protocol client and platform driver packages

### LINKED ISSUE #427: Split DevFlow Driver into portable protocol client and platform driver packages
## Summary

Split `Microsoft.Maui.DevFlow.Driver` into two packages with distinct compatibility and responsibility boundaries:

- **`Microsoft.Maui.DevFlow.Client`** targeting `netstandard2.0`, containing `AgentClient`, `ElementInfo`, protocol DTOs, and HTTP/JSON operations.
- **`Microsoft.Maui.DevFlow.Driver`** remaining modern and platform-specific, referencing `Microsoft.Maui.DevFlow.Client`, and containing native app process management, UI Automation, screenshots/Skia, recording, and other platform driver functionality.

This proposal intentionally does **not** retarget the full Driver package to .NET Standard.

## Motivation

The current Driver package targets `net9.0` only and combines the portable DevFlow wire protocol with platform-specific and modern runtime functionality. This prevents `net472` Visual Studio Apex tests from consuming the same DevFlow query and tap behavior used by .NET 10 `dotnet watch` and VS Code tests.

A portable protocol client would also enable a shared DevFlow scenario-session base across these test families without duplicating Agent HTTP/JSON protocol code.

## Current evidence

The package currently combines:

- `AgentClient`, `ElementInfo`, HTTP/JSON operations, and protocol DTOs
- Platform app drivers and process management
- Windows UI Automation
- Screenshots and Skia functionality
- Recording and newer platform/runtime APIs

The relevant dependencies—`System.Text.Json`, `SkiaSharp`, and `Interop.UIAutomationClient`—publish `netstandard2.0`-compatible assets. The split should nevertheless keep Skia, UI Automation, and other native/platform concerns in the modern Driver package rather than expanding the portable Client surface.

## Proposed design

1. Add `Microsoft.Maui.DevFlow.Client` targeting `netstandard2.0`.
2. Move `AgentClient`, element/protocol DTOs, serialization, and HTTP/JSON query/action operations into Client.
3. Keep `Microsoft.Maui.DevFlow.Driver` on modern target frameworks and reference Client.
4. Keep native process, platform UI, screenshot, Skia, and recording drivers in Driver.
5. Preserve existing Driver APIs for modern consumers where feasible, using forwarding/type-forwarding or compatible public surface organization as appropriate.
6. Define each protocol DTO once, in Client, so all test harnesses share serialization and behavior.

The custom `SocketsHttpHandler.ConnectCallback` and modern loopback connection implementation may require a conditional or alternative implementation for the `netstandard2.0` target.

## Acceptance criteria

- [ ] A `net472` consumer can reference `Microsoft.Maui.DevFlow.Client` and perform DevFlow query and tap operations.
- [ ] Existing `Microsoft.Maui.DevFlow.Driver` APIs and modern consumers continue working.
- [ ] Protocol DTOs are defined once and shared by Client and Driver consumers.
- [ ] Tests cover serialization plus query and tap behavior on both the .NET Framework/`netstandard2.0` consumer family and modern target family.
- [ ] The full Driver package is not retargeted to `netstandard2.0`; platform/native responsibilities remain in the modern package.

JanKrivanek:
Fixes #427

Splits `Microsoft.Maui.DevFlow.Driver` into a portable protocol client and a modern platform driver.

## What changed

- **New `Microsoft.Maui.DevFlow.Client`** (`netstandard2.0;net8.0`) — `AgentClient`, `ElementInfo`, the protocol DTOs and their serialization. This is what lets a `net472` Visual Studio Apex harness run the exact same DevFlow query/tap code as a .NET 10 `dotnet watch` or VS Code test run. Files were `git mv`'d, so the diff is rename-detected rather than a rewrite (`DriverJson` → `ProtocolJson` is the only rename inside).
- **`Microsoft.Maui.DevFlow.Driver`** stays modern and platform-specific (app process management, UI Automation, Mac accessibility, Skia, recording) and references Client. The namespace is unchanged and `TypeForwards.cs` re-exports every moved type.
- **netstandard2.0 support** — `SocketsHttpHandler.ConnectCallback` is `#if`-guarded. The portable target gets `LoopbackPreferenceHandler`, which races the IPv4/IPv6 loopback families with a probe socket, caches the winner, and pins the `Host` header so the request on the wire is identical to the modern path (the dotnet/maui-labs#341 behavior is preserved on both). Init-only/`required` polyfills and a minimal `NetStandardCompat` shim keep the DTO and client sources unforked.
- **`Microsoft.Maui.DevFlow.Client.Tests`** multi-targets `net472` (Windows) + `net10.0`, running the *same* source on both families: serialization/wire-shape, query/tree/element, tap/fill/`ActionResult`, and loopback family fallback.
- Solution files, `ci-cli.yml` path filters, package/architecture docs and a NuGet README updated. No pipeline change needed — the existing `Microsoft.Maui.DevFlow.*.nupkg` publish glob already covers the new package.

## Two things worth calling out

**The modern asset targets `net8.0`, not `net9.0`.** netstandard2.0 cannot carry `IsAotCompatible`/`IsTrimmable` (the SDK hard-errors), so with net9.0 as the only modern asset NuGet would hand .NET 8 (LTS) consumers the portable one — silently dropping the AOT/trim analyzers and the `SocketsHttpHandler` loopback dial from an AOT-capable app. Verified with a real net8.0 test project resolving `lib/netstandard2.0` before the change and `lib/net8.0` after. Nothing here needs a post-net8.0 API.

**`TreatAsLocalProperty` on the Client project.** The CLI's NativeAOT publish smoke test runs with `-p:PublishAot=true -p:PublishTrimmed=true`. MSBuild propagates those as global properties across the whole graph, and NuGet's restore-graph walk evaluates *every* target framework of *every* project in it — including the portable target, where the SDK rejects them (`NETSDK1207`). Global properties cannot be reassigned from a project file, hence `TreatAsLocalProperty` plus explicit opt-outs for netstandard2.0 only.

## Compatibility

**Source compatible.** The namespace is unchanged, so no consumer edits are needed. The CLI, integration tests and 405 existing DevFlow tests compile untouched.

**Binary compatible**, verified rather than assumed: I built the pre-split monolithic Driver from the base commit, compiled a consumer against it, then swapped in the new Driver + Client **without recompiling the consumer**. It runs correctly, with `AgentClient`/`ElementInfo` resolving from `Microsoft.Maui.DevFlow.Client` through the forwards and `AlertButton` still from Driver:

```
AgentClient          -> Microsoft.Maui.DevFlow.Client
ElementInfo          -> Microsoft.Maui.DevFlow.Client
AlertButton          -> Microsoft.Maui.DevFlow.Driver
GetStatusAsync       -> null (expected)
```

The forwards live entirely in our assembly; consumers need nothing. The one requirement is that `Microsoft.Maui.DevFlow.Client.dll` be resolvable at runtime, which the package dependency guarantees. It only bites someone hand-copying `Microsoft.Maui.DevFlow.Driver.dll` outside of NuGet restore.

## Validation

| Suite | Result |
|---|---|
| `Microsoft.Maui.DevFlow.Client.Tests` (net472) | 28/28 |
| `Microsoft.Maui.DevFlow.Client.Tests` (net10.0) | 28/28 |
| `Microsoft.Maui.DevFlow.Tests` | 405/405 |
| `Microsoft.Maui.Cli.UnitTests` | 805/805 |
| `DevFlow.slnf` build | clean |
| Old-consumer binary compat | passes without recompilation |

Packed both packages: Client ships `lib/netstandard2.0` + `lib/net8.0` (the latter carrying `IsTrimmable`), and Driver correctly declares its `Microsoft.Maui.DevFlow.Client` dependency.


src/DevFlow/Microsoft.Maui.DevFlow.Client/AgentClient.cs
```diff
@@ -144,6 +159,109 @@ private static async ValueTask<Stream> ConnectLoopbackAsync(
...
+                    request.Headers.Host ??= uri.IsDefaultPort ? uri.Host : $"{uri.Host}:{uri.Port}";
+                    request.RequestUri = new UriBuilder(uri) { Host = host }.Uri;
+                }
+            }
+
+            try
+            {
+                return await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
+            }
+            catch (HttpRequestException)
```
JanKrivanek, line 426: On .NET Framework, `HttpClientHandler` wraps a refused connection as `HttpRequestException -> WebException -> SocketException`. This rethrows into `IsTransientTransportException`, but that predicate only recognizes a `SocketException` when it is the *direct* inner exception; after recursing through `WebException`, the bare `SocketException` returns false. As a result, `TransientFailureRetryCount` is ignored on the new `netstandard2.0`/`net472` path, which is especially problematic when the client races agent or port-forward startup. Could we classify a bare `SocketException` as transient (or otherwise handle this chain) and run the existing delayed-listener retry scenario in the multi-targeted client tests?

JanKrivanek: Good catch — confirmed exactly as you described, and fixed in 057eeca0.

I reproduced the chain on both target families before changing anything:

| Target | Exception chain | tap, no retries | tap, 3 retries |
|---|---|---|---|
| net10.0 | `HttpRequestException -> SocketException` | 2062 ms | 9300 ms (retried) |
| net472 | `HttpRequestException -> WebException -> SocketException` | 6211 ms | 6073 ms (**never retried**) |

The recursion did reach the bare `SocketException`, but no `case` matched it, so it returned `false` and `TransientFailureRetryCount` was ignored on the portable path.

**Fix:** `IsTransientTransportException` now matches `SocketException` on its own, so it is transient at any depth. That subsumes the old `HttpRequestException { InnerException: SocketException }` case and covers the .NET Framework `WebException` chain. A socket failure in an HTTP send chain is a transport failure by definition, and retries stay opt-in via `TransientFailureRetryCount` (default 0).

**Tests** — new `AgentClientRetryTests`, running on both families:

- `TapAsync_RetriesTransientTransportFailures` — the real guard. It compares elapsed time against a permanently dead port with and without retries, so it depends only on the retry loop, not on how long a refusal takes on the host. Verified it fails without the fix (`net472: "-17 ms longer"`) and passes with it, while net10.0 passes either way.
- `TapAsync_SucceedsAgainstAnAgentThatStartsLate` and `GetStatusAsync_SucceedsAgainstAnAgentThatStartsLate` — the delayed-listener scenario you asked for, mirroring `WebViewDriverTests`.
- `TapAsync_MutatingRetriesCanBeDisabled` — `RetryMutatingRequests=false` still opts out.

Two notes from writing these. `GetStatusAsync` turned out to be a poor probe for this on its own: `GetStringWithRetryableUiReadAsync` gives it a ~1s retry window that masks whether transient retries fired at all, so the strict test uses `TapAsync`, which goes straight through `SendWithTransientRetriesAsync`. And the delta test needed a warm-up call, because the first refused connect on a host can be far slower than later ones and would otherwise inflate the baseline.

All suites green: Client tests 32/32 on net472 and net10.0, DevFlow 405/405.

src/DevFlow/Microsoft.Maui.DevFlow.Client.Tests/AgentClientQueryTests.cs
```diff
@@ -0,0 +1,148 @@
...
+    {
+        using var agent = FakeAgent.StartJson("[]");
+        using var client = new AgentClient("localhost", agent.Port);
+
+        await client.QueryAsync(type: "Button", automationId: "Submit Button", text: "a&b");
+
+        var request = Assert.Single(agent.Requests);
+        Assert.Equal("GET", request.Method);
+        Assert.Equal("/api/v1/ui/elements", request.Path);
+        Assert.Equal("type=Button&automationId=Submit%20Button&text=a%26b", request.Query);
```
jfversluis, line 73: Please add a net472/modern test for an escaped path segment such as `logs/app.txt`. .NET Framework may turn `%2F` back into `/`, changing the route used by file, preference, secure-storage, and action APIs.

src/DevFlow/Microsoft.Maui.DevFlow.Client.Tests/AgentClientRetryTests.cs
```diff
@@ -0,0 +1,141 @@
...
+        var retryDelay = TimeSpan.FromMilliseconds(400);
+
+        // Warm up first: on some hosts the very first refused connect is markedly slower than later
+        // ones, which would otherwise inflate the baseline and mask the retries.
+        await TimeTapAsync(port, retryCount: 0, retryDelay);
+
+        var withoutRetries = await TimeTapAsync(port, retryCount: 0, retryDelay);
+        var withRetries = await TimeTapAsync(port, retryCount: 2, retryDelay);
+
+        var difference = withRetries - withoutRetries;
```
jfversluis: Please assert retry attempt counts instead of comparing elapsed times. CI contention can delay either run and fail this test even when retry behavior is correct.

jfversluis (changes requested):
Two changes are needed before merging: preserve URL path behavior on net472 and make retry tests deterministic.

src/DevFlow/Microsoft.Maui.DevFlow.Client.Tests/AgentClientRetryTests.cs
JanKrivanek: Agreed — replaced the timing comparisons with attempt counts in 7ccb03ef.

The fake agent now resets each connection instead of answering, so every attempt still completes a TCP accept and gets recorded, which makes the count directly observable rather than inferred:

- `TapAsync_MakesExactlyOneAttemptPerConfiguredRetry` — `[Theory]` over 0/1/3 retries asserting exactly 1/2/4 attempts, with `TransientFailureRetryDelay = TimeSpan.Zero` so nothing depends on wall time.
- `TapAsync_StopsRetryingOnceTheAgentAnswers` — two failures then a real answer; asserts success and exactly 3 attempts, so a retry storm past the first success would fail.
- `TapAsync_DoesNotRetryWhenMutatingRetriesAreDisabled` — asserts exactly 1 attempt.

The classification itself is now covered directly, by exercising `IsTransientTransportException` (made `internal`, visible to the test assembly) with the precise chains each family produces. That is fully deterministic and pins the behaviour far more tightly than timing ever did. The two remaining delayed-listener tests assert only a boolean outcome, no elapsed time.

**Your comment also uncovered a second bug.** Once the tests counted attempts rather than milliseconds, net472 showed 1 attempt where it should have shown 4. A connection dropped mid-request — an agent restart, or an ADB port forward going away — is reported very differently per family:

| Target | Chain for a mid-request connection drop |
|---|---|
| net10.0 | `HttpRequestException -> HttpIOException` (an `IOException`, so it was already retried) |
| net472 | `HttpRequestException -> WebException(ConnectionClosed)` — **no inner exception at all** |

So neither the `SocketException` nor the `IOException` case saw it, and the request was never retried on .NET Framework. `WebException` is now classified by status, limited to transport-level ones (`ConnectFailure`, `ConnectionClosed`, `ReceiveFailure`, `SendFailure`, `KeepAliveFailure`, `NameResolutionFailure`, …). `ProtocolError` stays non-retryable because it is a real HTTP response the caller must see, and `Timeout`/`RequestCanceled`/`TrustFailure` stay non-retryable too — with tests asserting both directions.

The suite is also much faster now that nothing waits on backoff: 50 tests in 2 s (net10.0) and 9 s (net472), versus ~19 s and ~52 s before.

src/DevFlow/Microsoft.Maui.DevFlow.Client.Tests/AgentClientQueryTests.cs
JanKrivanek: Good instinct to check this — added `AgentClientPathEscapingTests` in 7ccb03ef, running on both target families.

I tested the concern before assuming either way, since the fake agent records the raw request line and can assert the exact bytes on the wire. **The good news: `%2F` survives intact on net472** with current defaults, so there is no production bug here. But it is very much worth pinning, because the failure mode would have been silent and portable-target-only.

Coverage, using `logs/app.txt` as you suggested:

- **Files** — `DownloadFileAsync`, `UploadFileAsync`, `DeleteFileAsync` all assert `/api/v1/storage/files/logs%2Fapp.txt`.
- **Preferences** — `GetPreferenceAsync` / `SetPreferenceAsync` with `group/setting`.
- **Secure storage** — `GetSecureStorageAsync` with `tokens/refresh`.
- **Actions** — `InvokeActionAsync` with `Area/Reset`.
- **Sensors** — `StartSensorAsync` with `orientation/raw`, checking the encoded segment stays intact ahead of the `/start` suffix.

Plus three edge cases worth locking down while I was in there:

- **Dot segments are not collapsed** — `logs/../secrets.txt` reaches the agent as `logs%2F..%2Fsecrets.txt`. This matters for more than routing: the agent owns the storage-root sandbox, so the client must not quietly rewrite a traversal-looking path into a different, possibly valid one.
- **Other reserved characters** — `my logs/a+b&c.txt` encodes to `my%20logs%2Fa%2Bb%26c.txt`, and importantly produces an empty query string, so a `&` in a filename cannot inject a query parameter.
- **Query strings stay separate** — `root=cache` remains a query parameter rather than being folded into the encoded path.

One caveat worth recording: this relies on the framework default. A consuming app can still opt into the legacy canonicalisation via `<uri><schemeSettings>` in `app.config`, which is outside our control — but these tests will catch it if the default itself ever shifts.

JanKrivanek (commented):
~~Dummy comment to hold off merging.
Doing last round review - then I'll resolve this~~

Reviewed

jfversluis (dismissed):
Re-reviewed the updates: encoded path behavior is now covered on both target families, retry tests assert deterministic attempt counts, and the additional net472 WebException classification is appropriately scoped. No remaining blocking findings.

jfversluis (approved):

PR #443, opened 2026-08-18
Simplify AppProjectReference for .NET 11

Redth:
## Why

.NET 11 Android and Apple SDKs now expose authoritative application outputs through `GetApplicationArtifacts` and `@(ApplicationArtifact)`. AppProjectReference currently duplicates that work by injecting child targets, overriding output paths, and scanning the file system for artifacts.

## What changed

- Add a .NET 11 path that projects `MauiAppProjectReference` into a standard `ProjectReference` using `ReferenceOutputAssembly="false"`, `BuildReference="true"`, `Targets="GetApplicationArtifacts"`, and `OutputItemType="MauiAppArtifact"`.
- Preserve the existing .NET 10-and-earlier implementation in separate legacy props/targets files.
- Keep Visual Studio builds on the legacy path because its project build manager does not honor `ProjectReference.Targets` the same way as command-line MSBuild.
- Restore each distinct app project once before reference resolution so clean consumers using the packed build package have child assets available.
- Preserve platform-owned artifact metadata and map compatibility fields such as `ArtifactType`, `PackageName`, `BundleIdentifier`, `Installable`, and `Launchable`.
- Document the standard MSBuild metadata and link to the authoritative Android, Apple, and MSBuild contracts.

The .NET 11 shorthand intentionally uses graph-visible `ProjectReference` metadata such as `SetTargetFramework`, `SetConfiguration`, and `AdditionalProperties`, avoiding a separate package-specific property model.

## Validation

- AppProjectReference test suite
- Static graph build coverage
- Clean packed-package consumer coverage
- AppProjectReference solution build


src/AppProjectReference/Microsoft.Maui.Build.AppProjectReference/build/Microsoft.Maui.Build.AppProjectReference.targets
```diff
@@ -1,198 +1,93 @@
...
-      <_MauiAppProjectReferenceForMSBuild Remove="@(_MauiAppProjectReferenceForMSBuild)" />
-      <_MauiAppProjectReferenceForMSBuild Include="@(_ResolvedMauiAppProjectReference)"
-                                        KeepMetadata="AdditionalProperties" />
-    </ItemGroup>
+    <RemoveDuplicates Inputs="@(_MauiAppProjectToRestore)">
+      <Output TaskParameter="Filtered" ItemName="_DistinctMauiAppProjectToRestore" />
+    </RemoveDuplicates>
 
-    <MSBuild Projects="@(_MauiAppProjectReferenceForMSBuild)"
+    <MSBuild Projects="@(_DistinctMauiAppProjectToRestore)"
```
jfversluis, line 56: This restore collapses references to project paths and then removes RID/TFM properties, so it loses `SetTargetFramework`, `SetConfiguration`, `SetPlatform`, and `AdditionalProperties`. A documented reference such as `SetTargetFramework="TargetFramework=net11.0-ios;RuntimeIdentifier=iossimulator-arm64"` is restored without the RID, but `ResolveProjectReferences` later invokes `GetApplicationArtifacts` with it. On a clean packed-package consumer, the child build can then fail with NETSDK1047 because `project.assets.json` has no target for that RID. Restore each distinct project/property combination and forward the standard `ProjectReference` metadata.

src/AppProjectReference/Microsoft.Maui.Build.AppProjectReference/build/Microsoft.Maui.Build.AppProjectReference.targets
```diff
@@ -1,198 +1,93 @@
...
-             Targets="_MauiAppRefBuildAndGetArtifacts"
-             BuildInParallel="$(BuildInParallel)"
-             SkipNonexistentProjects="false"
-             SkipNonexistentTargets="false">
-      <Output TaskParameter="TargetOutputs" ItemName="_MauiAppBuiltArtifact" />
-    </MSBuild>
+  <Target Name="_MauiAppRefCollectOnBuild"
+          BeforeTargets="$(MauiAppRefBuildBeforeTargets)"
+          DependsOnTargets="BuildAppProjectReferences"
+          Condition="'$(MauiAppRefUseProjectReferenceArtifacts)' == 'true'" />
```
jfversluis, line 68: The modern path always runs `BuildAppProjectReferences`; it ignores the still-public `MauiAppRefBuildEnabled` and `MauiAppRefBuildOnBuild` properties that the legacy path honors. Consumers that disable app builds—for example to avoid signing during a host build—will unexpectedly build the referenced app under .NET 11. Apply the existing enable/on-build conditions here, and preserve the corresponding VSTest behavior if `MauiAppRefBuildOnTest` remains supported.

jfversluis (changes requested):
The .NET 11 implementation has two compatibility issues that should be addressed before merging: the clean restore path drops reference-specific build properties, and the modern build hook no longer honors the existing opt-out controls.

PR #460, opened 2026-08-27
Add first-class Tizen platform identity to DevFlow contracts

Redth:
## Problem

DevFlow's public platform surface accepted `ios|android|maccatalyst|windows|linux|macos` and nothing else. The Tizen agent in [Redth/Maui.Tizen](https://github.com/Redth/Maui.Tizen) (validation PR #11) therefore had two bad options: report an identifier the schema rejects, or impersonate another platform.

Worse, it was already being impersonated for it. Tizen is a Linux distribution, so `OperatingSystem.IsLinux()` is true there — and every framework-neutral detection path (broker registration, the runtime profiler, the native agent fallback) tested Linux without testing Tizen first. A Tizen agent registered itself as **Linux**.

## What this adds

`tizen` is added **additively**. No value already on the wire changes: agents keep sending the spelling they always sent, and normalization is a client-side read concern.

### Canonical identity — `DevFlowPlatform`

New public static class in `Microsoft.Maui.DevFlow.Client` (type-forwarded from `Microsoft.Maui.DevFlow.Driver`, so already-compiled consumers keep binding). It is the single place mapping agent spellings (`iOS`, `MacCatalyst`, `WinUI`, `Tizen`) onto canonical lowercase ids, and it is deliberately forgiving:

- an identifier it does not recognize is **lowercased and returned unchanged**, never rejected and never coerced onto another platform — an older client stays usable against an agent on a newer platform. `IsKnown` distinguishes the two cases.
- `Matches` compares canonical identities but keeps the historical substring fallback, so existing partial filters (`--platform mac`) behave exactly as before.
- Tizen is probed **before** Linux everywhere it appears.

`AgentStatus` gains a `PlatformId` convenience (`JsonIgnore`, so the wire shape is untouched).

### Runtime detection — `DevFlowRuntimePlatform`

Framework-neutral detection in `Agent.Abstractions`, now the single implementation behind four previously duplicated copies (`DevFlowAgentService`, `DevFlowAgentHost`, `RuntimeProfilerCollector`, `Agent.Native`, plus the MAUI DI fallback). It probes the `TIZEN` OS platform, the runtime identifier, `/etc/tizen-release` and `Tizen.Applications.Common`, each independently guarded.

A `DEVFLOW_PLATFORM` environment variable overrides detection outright, so an out-of-tree agent on a platform DevFlow cannot yet detect never has to ship as `Unknown` — or as someone else — while waiting for detection support to land.

### Routing, not pretending

A canonical identifier does **not** imply a host-side driver. `AppDriverFactory.HasLocalDriver` makes that explicit:

- `AppDriverFactory.Create("tizen")` throws `PlatformNotSupportedException` with "recognized, but no local driver" rather than `ArgumentException: Unknown platform`.
- `maui devflow ui alert` ends in an iOS-simulator else-branch; a Tizen agent previously fell through to `xcrun`. It now fails with an explicit message.
- Agent filtering (`maui devflow agent wait --platform tizen`, `maui_list_agents`, `maui_wait`) matches canonical identity, so `tizen` finds an agent that registered as `Tizen` — and `linux` does not.
- Tizen never trips the Android adb forward/reverse plumbing: neither its platform name nor its `net10.0-tizen` TFM is mistaken for Android.

Everything over the agent protocol — tree, query, tap, screenshots, storage, logs — works normally on Tizen. Only host-side driving (launch, record, alerts, system-scope theme) is unavailable, and it now says so.

### MCP

`maui_list_agents` and `maui_wait` gain an optional, alias-aware `platform` filter (last parameter, defaulted — existing callers unaffected) and surface `platformId` next to the agent-reported spelling.

### Schema and docs

`agent-status.json` gains `tizen`; `device.json` descriptions updated. The spec README gains a **Platform identity** section covering the canonical table, the pass-through rule, external agent usage and the minimum package version. `broker.md`, `src/DevFlow/README.md` and the client README document the identity-vs-driver split.

## Tests

63 new tests, all passing:

- **`DevFlowPlatformTests`** (Client) — parse/normalize across spellings, the explicit "Tizen is not Linux" assertion, unknown-platform pass-through, display names, filter matching, and `AgentStatus` round-trip proving the wire value is preserved verbatim. Runs on both the portable and modern target families.
- **`TizenPlatformIdentityTests`** (DevFlow) — the `DEVFLOW_PLATFORM` override, driver-factory recognition plus every pre-existing alias still resolving, and a schema test asserting the enum is additive **and** exactly equal to `DevFlowPlatform.KnownIds`, so the schema and client cannot drift.
- **`DevFlowTizenPlatformTests`** (CLI) — discovery/filtering, the `maui devflow diagnose --json` command path asserting a Tizen agent is reported without triggering Android probing, the alert-command guard, the MCP parameter contract, and registration round-trip.

## Validation

| Suite | Result |
|---|---|
| `Microsoft.Maui.DevFlow.Client.Tests` | 93 passed |
| `Microsoft.Maui.Cli.UnitTests` | 884 passed |
| `Microsoft.Maui.DevFlow.Tests` (Tizen + ProtocolSpec) | 28 passed |
| `dotnet build src/DevFlow/DevFlow.slnf` | 0 errors |
| `dotnet pack` Client + Agent.Abstractions | packages produced |

The full `Microsoft.Maui.DevFlow.Tests` run has 7 failures on this machine; all 7 reproduce on `main` unmodified and are environmental (loopback `SocketException: Can't assign requested address`, a path-separator assumption, and analyzer-host timeouts). None touch platform identity.

## For Maui.Tizen

With `Microsoft.Maui.DevFlow.* 0.1.0-preview.13`, **the platform-name workaround can be deleted outright**. A MAUI Tizen backend gets the correct name for free — `Agent.Core` reports `DeviceInfo.Current.Platform.ToString()`, which is already `Tizen` — and the framework-neutral broker registration path no longer reports Linux. No spoofing, no local shim, no `DEVFLOW_PLATFORM` needed.

## Behavior changes

Additive in substance: a new type, a new `JsonIgnore` property, new optional trailing parameters, and a widened schema enum. Three deliberate behavior changes are worth calling out explicitly:

1. **Agent filtering is stricter when both sides are known platforms.** Filtering previously used a raw substring match. It now compares canonical identity first, and only falls back to substring when one side is unrecognized — otherwise a `linux` filter would still match an agent reporting `Tizen (Linux) 8.0`, which is the bug this PR exists to fix. Concretely, `--platform mac` is an alias for Mac Catalyst and no longer also matches macOS agents. Partial filters such as `tiz` and `andro` are unaffected.
2. **`AppDriverFactory.Create` throws `PlatformNotSupportedException`** for `tizen` and `macos`, where it previously threw `ArgumentException` — those strings were indistinguishable from an unknown platform before, so no caller could have depended on the old type. Documented via `<exception>` on the method.
3. **Mac Catalyst reports `MacCatalyst` in the profiler/host-identity fallback**, where one of the four now-deduplicated detection copies tested `IsIOS()` before `IsMacCatalyst()` and so reported `iOS`. This is wire-visible, and is a correctness fix consistent with the other three copies.


Redth:
## Review findings addressed — `6e4011b`

Thanks — the consensus findings were correct, and finding #1 was a genuine hole in the central fix rather than a nit.

### 1. 🟡 `Matches` substring fallback — **fixed**

This was a real bug and it undermined the whole point of the PR. An agent reporting a decorated string such as `"Tizen (Linux) 8.0"` normalizes to `tizen`, but the substring fallback then let a `linux` filter match it anyway — reintroducing exactly the Tizen/Linux ambiguity this change exists to remove.

Canonical equality is now authoritative whenever **both** sides resolve to platforms DevFlow knows. The substring fallback survives only when one side is unrecognized, so partial filters (`tiz`, `andro`, raw TFM fragments) keep matching as before.

One deliberate consequence: `--platform mac` is an alias for Mac Catalyst and no longer also sweeps up macOS agents. That is the correct semantics, and it is now asserted rather than incidental. New tests cover decorated Tizen strings against `linux`/`gtk` filters at both the `DevFlowPlatform` and CLI-discovery layers.

### 2. 🟡 `DEVFLOW_PLATFORM` override — **partially fixed, precedence kept by design**

The sanitization half was right and is fixed: the value goes verbatim onto the wire and into JSON, so it is now validated before use (≤32 chars; letters, digits, `.`, `-`, `_`, space). Anything else is ignored in favour of detection. Tests cover JSON-breaking, header-injection, path-traversal and over-length values.

I kept unconditional precedence over detection, and documented why. The failure this exists to solve is detection returning a confidently **wrong** answer, not detection giving up: a Tizen host satisfies `OperatingSystem.IsLinux()`, so an override that only applied when the result was `Unknown` would never get the chance to correct it — which is the original bug.

### 3. 🟢 Exception-type change — **documented**

`<exception>` docs on `AppDriverFactory.Create` now state that `PlatformNotSupportedException` replaces `ArgumentException` for recognized-but-undriveable platforms, and the PR description carries the same note.

### Also fixed (raised by a single reviewer, but correct)

- **`Create`'s "Supported:" list advertised `macos`/`tizen`**, which it cannot construct. It now lists only constructible platforms — and `macos` joins `tizen` in failing with the explicit `PlatformNotSupportedException` rather than a misleading "Unknown platform". There is no AppKit host driver; `MacCatalystAppDriver` covers Mac Catalyst only.
- **`ResolveAlertPlatformAsync`** used raw `.Contains(...)`; it now normalizes, consistent with the rest of the PR.
- **Bare `"nui"` alias dropped** as too broad. `tizen-nui`/`tizennui` remain.
- **Env-mutating tests** moved into an `EnvironmentVariables` collection to stop them racing other classes.
- **Schema enum vs. pass-through contradiction** — resolved by stating the asymmetry outright instead of papering over it: the `enum` is a closed, additively grown set of platforms DevFlow ships identity for, while clients are lenient at runtime. An agent on an unlisted platform therefore *works* but will not *validate* until its identifier is added. Both the schema description and the spec README now say so.

### Not changed, with reasoning

- **`ResolveAlertPlatformAsync` host fallback lacks a Tizen branch.** Intentional, now commented. That branch describes the machine running the CLI, and `maui` is a .NET SDK global tool that runs on a developer host — never on a Tizen device. Tizen is only ever reached as a remote agent, which the normalized agent-status path above handles.
- **`IsTizen` ignores the override.** By design, now documented: it reports what the process is actually running on and is used for ordering, whereas `DetectName` produces the reported name. The divergence is the point.
- **`tvOS` absent from the enum.** Pre-existing and out of scope. Worth noting `Normalize("tvOS")` correctly yields `tvos` rather than being swallowed by the `ios` token.
- **`FindSpecRoot()` filesystem walk.** Mirrors the existing pattern in `ProtocolSpecTests`; changing it is unrelated churn.

### Validation

| Suite | Result |
|---|---|
| `Microsoft.Maui.DevFlow.Client.Tests` | 100 passed |
| `Microsoft.Maui.Cli.UnitTests` | 885 passed |
| `Microsoft.Maui.DevFlow.Tests` (Tizen + ProtocolSpec) | 37 passed |

70 platform-identity tests total, up from 63.

Redth:
POC as we look at moving Tizen to its own repo.  No intention of merging this currently.

PR #488, opened 2026-09-10
Make pinch, pan and swipe reach controls that own their touch handling

aritchie:
Pinch, pan and swipe were already wired end to end (`maui_gesture` → `/ui/actions/gesture` → recognizer tier → native tier). What was missing was the native tier on iOS, Mac Catalyst and Android — the cases where a control owns its own touch handling.

## What was broken

**iOS / Mac Catalyst**
- `pan` only moved a `UIScrollView`'s content offset. A map, a canvas, or a custom drag surface returned "no handler".
- `swipe` had no native path at all — it fell back to `ScrollToAsync`.
- Naive recognizer driving reports success while doing nothing: it walks up the tree and drives the navigation controller's `UIScreenEdgePanGestureRecognizer`, or a `UIScrollView`'s own pan recognizer, which UIKit ignores because scroll views scroll from their internal touch tracking.
- Nothing reached views that take touches directly and install no recognizer, such as MAUI's `GraphicsView`.

**Android**
- Any *vertical* drag on an element inside a `ScrollView` was stolen by the scroll view one move event in: a requested `dy=60` arrived as `dy=6`. Horizontal drags worked, which made it look intermittent.
- Pans overshot by ~50% (`-100` scrolled `-150`) because lifting straight after the last move leaves velocity on the `VelocityTracker` and the view flings past the target.

## What changed

**iOS / Mac Catalyst pan** — MKMapView re-centre → scroll view (only when it can actually move) → the element's **own** `UIPanGestureRecognizer` (`SetTranslation` is public API) → synthetic touches → an ancestor's pan recognizer last. A scroll view's built-in pan (and UIKit's private pans on it) and screen-edge recognizers are excluded, so a gesture no longer reports success for something the platform silently ignored; a custom pan an app attaches to a scroll view is still driven.

**iOS / Mac Catalyst swipe** — a real native path: `UISwipeGestureRecognizer` when one matches the direction, otherwise a fast pan over the same surfaces.

**Raw-touch tier** (`AppleSyntheticTouch.cs`, new) — synthesised `UITouch`es delivered straight to a view. Notes on the mechanics:
- **Who gets them is an allow-list, not a heuristic.** Overriding `touchesBegan:` is also true of `UIControl`, `UIScrollView` and text views, so `AppleRawTouchPolicy` accepts MAUI's `GraphicsView` plus any type listed in `AgentOptions.SyntheticTouchViewTypes`. `UIControl`/`UIScrollView` subclasses and views with an enabled gesture recognizer are refused whatever is listed.
- **Every finger must land on the resolved view.** The target is resolved once — it must be the named element or inside it — and the injector refuses the sequence if any starting point hit-tests elsewhere, so an overlay never receives touches the response attributes to the canvas.
- **Geometry stays on the view** (`SyntheticTouchGeometry`, unit-tested): pinches from an edge origin move inwards, and a drag longer than the view is shortened along its own direction — the response detail says so.
- `UITouch` binds its target through `_responder`/`_cachedResponderView` on current UIKit, not the `_view` that older write-ups assume.
- The event is a `UIEvent` **subclass** overriding `AllTouches`/`TouchesForView` — public API, and required because MAUI's `GraphicsView` reads the event rather than the touch set it is handed.
- Every ivar write checks the type encoding first, and the finished touch is verified through the public API, so an OS change degrades to "no handler" instead of delivering a nonsense touch.
- Off by default behind `AgentOptions.EnableSyntheticTouch`. Advertised as `capabilities.syntheticTouch` in `/agent/status` — true only on iOS/Mac Catalyst with the option on — and exposed on the client's `AgentCapabilities` and in `agent-status.json`.
- SkiaSharp's `SKCanvasView` is **not** reached by this tier: its `SKTouchHandler` is a `UIGestureRecognizer`, so touches handed to the view never reach it.

**Android** — `RequestDisallowInterceptTouchEvent` on ancestors for the duration of an injected gesture, so a gesture aimed at a named element stays there (the target view still handles it normally); and a settle-before-lift on pans so they stop where asked. Swipes deliberately keep their fling.

## Verification

Run against real apps, asserting what the app *received* rather than an HTTP 200:

| | iOS sim | Mac Catalyst | Android emu |
|---|---|---|---|
| recognizer tier (pinch/pan/swipe/doubletap) | ✅ | ✅ | ✅ |
| native scroll tier | ✅ | ✅ | ✅ (content moved, checked via element bounds) |
| raw-touch canvas: pinch | ✅ `maxtouches=2 scale=0.50` | ✅ | ✅ |
| raw-touch canvas: pan/swipe, incl. vertical inside a ScrollView | ✅ | ✅ | ✅ |

New coverage after review:
- Integration (iOS/Mac Catalyst): a pan on an ordinary `Button` never presses it; raw swipe; pinch from `(0,0)` and `(1,1)` origins; a 5000pt pan on the canvas is shortened and still delivered; `capabilities.syntheticTouch` is true only on UIKit platforms.
- Unit: `SyntheticTouchGeometryTests` (edge origins, oversized and diagonal drags, invalid input), status capability reporting per backend, and client deserialisation of `syntheticTouch`.

After the review fixes, `UiActionTests` + `AgentMetadataTests` are 51/51 on Mac Catalyst (iOS simulator and Android not re-run for this round). Unit: `SyntheticTouchGeometryTests`, lifecycle and capability tests 82/82; `Client.Tests` 56/56.

## Out of scope

Windows, macOS AppKit and the non-MAUI native agent are unchanged; the native agent still rejects pinch and rotate outright.

Noticed while verifying, not addressed here: MAUI's `ScrollY` property does not update after a native scroll on Android — the content moves, the property does not.

https://claude.ai/code/session_01ECqr5aLjftZ98gokaDUPiB

### DIFF
--- a/src/DevFlow/Microsoft.Maui.DevFlow.Agent/AppleSyntheticTouch.cs
+++ b/src/DevFlow/Microsoft.Maui.DevFlow.Agent/AppleSyntheticTouch.cs
@@ -1,0 +1,390 @@
+#if IOS || MACCATALYST
+using System.Runtime.InteropServices;
+using CoreGraphics;
+using Foundation;
+using ObjCRuntime;
+using UIKit;
+
+namespace Microsoft.Maui.DevFlow.Agent;
+
+/// <summary>
+/// A synthesised <see cref="UITouch"/>, for the views that recogniser driving cannot reach:
+/// SkiaSharp's SKCanvasView, MAUI's GraphicsView, and any other UIView that takes input by
+/// overriding <c>touchesBegan:</c> instead of installing a gesture recogniser. Those views own
+/// no recogniser to drive and no scroll offset to nudge, so a synthetic touch is the only way in.
+///
+/// UITouch exposes no public way to set a location or a phase, so both are written through the
+/// object's ivars. Every write checks the ivar's type encoding first, and the finished touch is
+/// verified through the public API before it is handed to anyone — if UIKit's layout is ever not
+/// what we assume, <see cref="Create"/> returns null and the gesture reports "no handler" rather
+/// than delivering a nonsense touch. Off by default; opt in with
+/// <c>AgentOptions.EnableSyntheticTouch</c>.
+/// </summary>
+internal sealed class AppleSyntheticTouch : IDisposable
+{
+    [DllImport("/usr/lib/libobjc.dylib")]
+    internal static extern IntPtr object_getClass(IntPtr obj);
+
+    [DllImport("/usr/lib/libobjc.dylib")]
+    internal static extern IntPtr class_getMethodImplementation(IntPtr cls, IntPtr selector);
+
+    [DllImport("/usr/lib/libobjc.dylib")]
+    private static extern IntPtr class_getInstanceVariable(IntPtr cls, string name);
+
+    [DllImport("/usr/lib/libobjc.dylib")]
+    [return: MarshalAs(UnmanagedType.I1)]
+    private static extern bool class_respondsToSelector(IntPtr cls, IntPtr selector);
+
+    [DllImport("/usr/lib/libobjc.dylib")]
+    private static extern nint ivar_getOffset(IntPtr ivar);
+
+    [DllImport("/usr/lib/libobjc.dylib")]
+    private static extern IntPtr ivar_getTypeEncoding(IntPtr ivar);
+
+    [DllImport("/usr/lib/libobjc.dylib", EntryPoint = "objc_msgSend")]
+    private static extern IntPtr SendMessage(IntPtr receiver, IntPtr selector);
+
+    [DllImport("/usr/lib/libobjc.dylib", EntryPoint = "objc_msgSend")]
+    private static extern void SendMessage(IntPtr receiver, IntPtr selector, nint argument);
+
+    /// <summary>alloc/init by class name, for the UIKit types with no usable public constructor.</summary>
+    internal static IntPtr Allocate(string className)
+    {
+        var cls = (IntPtr)Class.GetHandle(className);
+        if (cls == IntPtr.Zero)
+            return IntPtr.Zero;
+
+        var handle = SendMessage(cls, Selector.GetHandle("alloc"));
+        return handle == IntPtr.Zero ? IntPtr.Zero : SendMessage(handle, Selector.GetHandle("init"));
+    }
+
+    private CGPoint _location;
+
+    private AppleSyntheticTouch(UITouch touch, CGPoint location)
+    {
+        Touch = touch;
+        _location = location;
+    }
+
+    internal UITouch Touch { get; }
+
+    /// <summary>
+    /// Builds a touch that reports <paramref name="locationInWindow"/> and starts in the
+    /// Began phase, or null when the touch cannot be assembled convincingly.
+    /// </summary>
+    internal static AppleSyntheticTouch? Create(UIWindow window, UIView view, CGPoint locationInWindow)
+    {
+        var handle = Allocate("UITouch");
+        if (handle == IntPtr.Zero)
+            return null;
+
+        // owns: true — the +1 from alloc becomes the wrapper's, so disposing deallocates.
+        var touch = Runtime.GetNSObject<UITouch>(handle, owns: true);
+        if (touch == null)
+            return null;
+
+        var synthetic = new AppleSyntheticTouch(touch, locationInWindow);
+        if (synthetic.Configure(window, view, locationInWindow))
+            return synthetic;
+
+        synthetic.Dispose();
+        return null;
+    }
+
+    private bool Configure(UIWindow window, UIView view, CGPoint location)
+    {
+        var handle = (IntPtr)Touch.Handle;
+
+        // These are strong ivars — the touch releases them when it deallocates, so they are
+        // retained on the way in to keep the counts balanced.
+        if (!TryWriteObject(handle, "_window", (IntPtr)window.Handle)
+            || !TryBindView(handle, view)
+            || !TryWritePoint(handle, "_locationInWindow", location)
+            || !TrySetPhase(UITouchPhase.Began))
+            return false;
+
+        // Best-effort extras: a handler that reads them gets sensible values, and a
+        // handler that does not is unaffected if the ivar has moved.
+        TryWritePoint(handle, "_previousLocationInWindow", location);
+        TryWriteInteger(handle, "_tapCount", 1);
+        TryWriteDouble(handle, "_timestamp", NSProcessInfo.ProcessInfo.SystemUptime);
+
+        return Verify(window, location, UITouchPhase.Began);
+    }
+
+    /// <summary>
+    /// Points the touch at the view that should receive it. UIKit has held this under several
+    /// names — a plain <c>_view</c> on older releases, a <c>_responder</c> plus a cached view on
+    /// current ones — so each is attempted and one hit is enough.
+    /// </summary>
+    private static bool TryBindView(IntPtr handle, UIView view)
+    {
+        var target = (IntPtr)view.Handle;
+        var bound = TryWriteObject(handle, "_view", target);
+        bound |= TryWriteObject(handle, "_responder", target);
+        bound |= TryWriteObject(handle, "_cachedResponderView", target);
+        return bound;
+    }
+
+    /// <summary>Reads the touch back through the public API to confirm the ivar writes landed.</summary>
+    private bool Verify(UIView reference, CGPoint expected, UITouchPhase phase)
+    {
+        try
+        {
+            var actual = Touch.LocationInView(reference);
+            return Math.Abs(actual.X - expected.X) < 1
+                && Math.Abs(actual.Y - expected.Y) < 1
+                && Touch.Phase == phase;
+        }
+        catch (Exception ex)
+        {
+            System.Diagnostics.Debug.WriteLine(
+                $"[Microsoft.Maui.DevFlow] Synthetic touch verification failed: {ex.GetBaseException().Message}");
+            return false;
+        }
+    }
+
+    internal bool MoveTo(CGPoint locationInWindow)
+    {
+        var handle = (IntPtr)Touch.Handle;
+        TryWritePoint(handle, "_previousLocationInWindow", _location);
+        if (!TryWritePoint(handle, "_locationInWindow", locationInWindow))
+            return false;
+
+        TryWriteDouble(handle, "_timestamp", NSProcessInfo.ProcessInfo.SystemUptime);
+        _location = locationInWindow;
+        return TrySetPhase(UITouchPhase.Moved);
+    }
+
+    internal bool End() => TrySetPhase(UITouchPhase.Ended);
+
+    private bool TrySetPhase(UITouchPhase phase)
+    {
+        var handle = (IntPtr)Touch.Handle;
+        if (TryWriteInteger(handle, "_phase", (long)phase))
+            return true;
+
+        // Some runtimes keep the phase behind a private setter rather than a plain ivar.
+        var selector = (IntPtr)Selector.GetHandle("setPhase:");
+        if (class_respondsToSelector(object_getClass(handle), selector))
+        {
+            SendMessage(handle, selector, (nint)(long)phase);
+            return true;
+        }
+
+        return false;
+    }
+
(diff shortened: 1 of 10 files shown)


src/DevFlow/Microsoft.Maui.DevFlow.Agent/AppleSyntheticTouch.cs
```diff
@@ -0,0 +1,390 @@
...
+    {
+        try
+        {
+            var selector = (IntPtr)Selector.GetHandle("touchesBegan:withEvent:");
+            var stock = AppleSyntheticTouch.class_getMethodImplementation(
+                (IntPtr)Class.GetHandle("UIView"), selector);
+            var actual = AppleSyntheticTouch.class_getMethodImplementation(
+                AppleSyntheticTouch.object_getClass((IntPtr)view.Handle), selector);
+            return stock != IntPtr.Zero && actual != IntPtr.Zero && actual != stock;
+        }
```
jfversluis, line 451: `touchesBegan:withEvent:` is not a reliable signal for a raw canvas. Common UIKit controls—including `UIControl`/`UIButton`, `UIScrollView`, text fields, switches, sliders, and collection/table views—override it too. With synthetic touch enabled, this can deliver a pan sequence to a button and fire its action, or report native success on a boundary scroll view without moving anything. Please replace the IMP-difference heuristic with a conservative positive allow-list/registration mechanism and explicitly exclude ordinary controls, scroll views, and recognizer-backed views.

src/DevFlow/Microsoft.Maui.DevFlow.Agent/AppleSyntheticTouch.cs
```diff
@@ -0,0 +1,390 @@
...
+        if (window == null)
+            return false;
+
+        var start = positionsAt(0);
+        if (start.Length == 0)
+            return false;
+
+        // Everything is delivered to the view under the first finger, the way UIKit keeps a
+        // gesture with the view that received its first touch.
+        var target = window.HitTest(start[0], null) ?? view;
```
jfversluis: `AppleSyntheticPinchAsync` validates the raw-touch target using the gesture center, but the injector hit-tests again here using the first finger's start point. An overlay under that finger can therefore receive both touches while the response names the center canvas. Pass the already resolved target into `InjectAsync` and validate that every initial touch point belongs to that same target instead of resolving a second target here.

src/DevFlow/Microsoft.Maui.DevFlow.Agent/PlatformAgentService.Gestures.cs
```diff
@@ -596,23 +714,144 @@ private static bool TrySetState(UIGestureRecognizer recognizer, UIGestureRecogni
...
+    /// middle. Both calls deal in CLLocationCoordinate2D, which stays reachable by reflection
+    /// because the value is only ever handed straight back to MapKit, never picked apart.
+    /// </summary>
+    private static string? TryPanMapView(UIView mapView, double deltaX, double deltaY)
+    {
+        try
+        {
+            var mapType = mapView.GetType();
+            var convert = mapType.GetMethod(
+                "ConvertPoint", [typeof(CoreGraphics.CGPoint), typeof(UIView)]);
```
jfversluis, line 841: The Xamarin/UIKit binding exposes this conversion as `ConvertPointToView` (and this file calls that method directly at line 630), so reflecting `ConvertPoint` returns null and the new MKMapView recenter path exits before `SetCenterCoordinate`. Reflect `ConvertPointToView`, or avoid reflection for the point conversion if the map type can be handled directly.

jfversluis (changes requested):
Requesting changes because the new Apple raw-touch target selection can dispatch synthetic touches to ordinary UIKit controls rather than only direct-touch canvases. That can turn a requested pan into a `UIControl` action and can reintroduce false-success responses for `UIScrollView`.

Please address the remaining blockers and related correctness gaps before merging:

- make raw-touch eligibility conservative (an explicit allow-list/registration contract, with ordinary controls and scroll views excluded);
- pass the already validated target into the injector and ensure every initial touch point belongs to that target;
- platform-gate `capabilities.syntheticTouch` and add it to both the public `AgentCapabilities` DTO and `agent-status.json`;
- clamp or explicitly reject raw drags that start outside the target, and reject only a scroll view's built-in pan recognizer rather than every custom pan attached to a scroll view;
- add negative/boundary coverage for ordinary UIKit controls, raw swipe, edge origins, oversized deltas, tier selection, and capability serialization.

Correction: the earlier MKMapView reflection concern was withdrawn after verifying that `MapKit.MKMapView.ConvertPoint(CGPoint, UIView)` is the correct binding for `convertPoint:toCoordinateFromView:` and returns `CLLocationCoordinate2D`.

The Android interception/settling changes and the defensive UIKit ivar verification are strong; the blocking problems are in Apple raw-touch target routing and the surrounding capability contract.

src/DevFlow/Microsoft.Maui.DevFlow.Agent/PlatformAgentService.Gestures.cs
aritchie: I checked this against the bindings. The reflection as written is correct — `ConvertPoint` here is MKMapView's own method, not UIView's point-to-point conversion.

Dumping `Microsoft.iOS.dll` (iOS 26.5 ref pack):

- `MapKit.MKMapView.ConvertPoint(CGPoint, UIView)` — the binding for `convertPoint:toCoordinateFromView:`, returns `CLLocationCoordinate2D`
- `UIKit.UIView.ConvertPointToView(CGPoint, UIView)` — `convertPoint:toView:`, returns `CGPoint`

Line 630 uses the second one because it genuinely wants a window point. The MKMapView path needs a coordinate to hand to `SetCenterCoordinate`, so reflecting `ConvertPointToView` would return a `CGPoint` and the `SetCenterCoordinate` invoke would throw on the argument type. Leaving this one as is.

jfversluis: You're right. I verified this against the official .NET iOS documentation and the installed iOS/Mac Catalyst 26.5 reference packs: `MapKit.MKMapView.ConvertPoint(CGPoint, UIView)` maps `convertPoint:toCoordinateFromView:` and returns `CLLocationCoordinate2D`, while `UIKit.UIView.ConvertPointToView` is the separate point-to-view conversion. I’m withdrawing this finding and resolving the thread.

src/DevFlow/Microsoft.Maui.DevFlow.Agent/AppleSyntheticTouch.cs
aritchie: Agreed, fixed in b0dc7488. The IMP-difference check is gone. `AppleRawTouchPolicy` now decides acceptance with a positive allow-list:

- Accepted: MAUI's `PlatformTouchGraphicsView`, plus types the app names in the new `AgentOptions.SyntheticTouchViewTypes` (simple or full type name; subclasses match; registering `UIView` itself matches nothing).
- Always refused, even if registered: `UIControl` and `UIScrollView` subclasses (so buttons, switches, sliders, text fields, text/table/collection views), and any view with an enabled gesture recognizer other than hover. GraphicsView installs only a `UIHoverGestureRecognizer`.
- The `touchesBegan:` override is still required, but only as a sanity check on registered types, so touches aren't forwarded up the responder chain.

This also corrected a claim in the PR: SkiaSharp's `SKCanvasView` is not reachable this way. Its `SKTouchHandler` is a `UIGestureRecognizer`, so touches handed to the view never reach it. The recognizer exclusion refuses it, and the docs no longer mention it.

Coverage: `Pan_OnAnOrdinaryButton_NeverDeliversSyntheticTouches` pans a plain `Button` on the gesture page and asserts it saw no Pressed or Clicked. It passes on Mac Catalyst.

src/DevFlow/Microsoft.Maui.DevFlow.Agent/AppleSyntheticTouch.cs
aritchie: Fixed in b0dc7488. The target is now resolved once and passed in:

- `ResolveRawTouchTarget` hit-tests the element's centre. The hit view must be the named element or a descendant of it, and must pass the policy.
- `AppleTouchInjector.InjectAsync(target, ...)` no longer hit-tests on its own and no longer falls back to `?? view`. Before delivering anything it checks that **every** finger's starting point hit-tests to that exact target, and it reports unhandled on any mismatch. The response therefore always names the view that actually received the touches.
- Pinch and drag geometry now come from the resolved view's bounds rather than the element's (`SyntheticTouchGeometry`), so both fingers start on it. Edge origins are moved inwards, and drags longer than the view are shortened along their direction, with the response detail noting it.

Coverage: unit tests for edge origins, oversized and diagonal drags, and invalid input. Mac Catalyst integration tests for pinch from `(0,0)` and `(1,1)` and a 5000pt pan also pass.

src/DevFlow/Microsoft.Maui.DevFlow.Agent/PlatformAgentService.Gestures.cs
```diff
@@ -258,6 +263,12 @@ bool Send(MotionEventActions action, int activePointers)
                     (MotionEventActions)((int)MotionEventActions.PointerDown | (i << PointerIndexShift)),
                     i + 1);
 
+            // Keep the gesture on the element the caller named. Without this an enclosing
+            // ScrollView steals a vertical drag once it passes the touch slop — correct for a
+            // real finger, wrong for "pan this canvas". Ancestors only: the target view itself
+            // still handles the gesture normally. Must follow the down, which resets the flag.
+            targetView.Parent?.RequestDisallowInterceptTouchEvent(true);
+
```
jfversluis, line 271: This still does not guarantee that the named element owns the gesture. `RequestDisallowInterceptTouchEvent(true)` runs only after `ACTION_DOWN`/pointer-down has already been dispatched. Android geometry can put that initial point outside the target: `GetAndroidTouchGeometry` keeps a fixed radius around `originX=0/1`, and `AndroidPanAsync` starts at `center - delta/2` without shortening an oversized delta. `activity.DispatchTouchEvent` can therefore select a sibling or ancestor, return `true`, and report native success for the requested element while the other view received the stream. Please apply the bounds-safe geometry in pixel units, or reject/shorten requests so every initial pointer lands inside the target, and add Android edge-origin and oversized-pan coverage.

src/DevFlow/Microsoft.Maui.DevFlow.Agent/AppleSyntheticTouch.cs
jfversluis: Confirmed in b0dc7488. The positive allow-list, hard exclusions for controls/scroll views/recognizer-backed views, and the ordinary-button integration test address this finding.

src/DevFlow/Microsoft.Maui.DevFlow.Agent/AppleSyntheticTouch.cs
jfversluis: Confirmed in b0dc7488. The target is resolved once, passed into the injector, and every initial pointer must hit-test to that exact view; the edge-origin and oversized-drag coverage exercises the corrected UIKit path.

src/DevFlow/Microsoft.Maui.DevFlow.Agent/PlatformAgentService.Gestures.cs
aritchie: Fixed in c59c846a, using the same rules as the Apple path:

- **Pixel-safe geometry.** Android now uses `SyntheticTouchGeometry`. It works in dp across the view's visible rect, then converts to pixels. Edge-origin pinches and rotations move inwards (`ForRotation` is new and has unit tests). Pans and swipes longer than the view are shortened along their direction, and the detail says so.
- **Ownership.** Events go to the element's view (its `ContainerView` when wrapped) rather than `activity.DispatchTouchEvent`. A sibling or intercepting ancestor can no longer take the stream while the response names the element. If the view refuses `ACTION_DOWN`, the gesture reports unhandled.
- **Visibility.** A view that is partly or fully scrolled out of sight is scrolled on screen before it is touched. Pointers are placed only inside its visible rect.

Coverage: the raw swipe, edge-origin pinch `(0,0)`/`(1,1)` and 5000dp pan integration tests now run on Android too. The full `UiActionTests` suite passes on an Android emulator (42/42).

PR #491, opened 2026-09-14
Enable reliable startup profiling injection on iOS

simonrozsival:
## Summary

Make startup profiling build injection reliable for projects that disable implicit usings and enable the supported iOS profiling path.

## Root cause

The injected bootstrap source referenced `TimeSpan` and `DateTime` without importing `System`, so it failed to compile in consumer projects with implicit usings disabled. The profiling session setup also unconditionally suppressed build injection for iOS, and the injection targets only supplied helper environment variables through Android environment files.

## Changes

- Add `using System;` to the injected profiling bootstrap source.
- Remove the unconditional iOS build-injection suppression.
- Add iOS `MlaunchEnvironmentVariables` entries for `MAUI_PROFILING_HELPER`, `MAUI_PROFILING_HELPER_EXIT_HOST`, and `MAUI_PROFILING_HELPER_EXIT_PORT`.
- Add a focused regression test covering the self-contained source import and iOS launch environment entries.

## Scope boundaries

- Android EventPipe and runtime PGO environment generation remains Android-only.
- Mac Catalyst support is explicitly out of scope; the profiling command continues to reject Mac Catalyst.
- This bottom-layer PR excludes Android dsrouter changes, trace-stop timeout changes, Windows dnx changes, and unrelated documentation updates from the source snapshot.

## Testing

- `./eng/common/dotnet.sh test src/Cli/Microsoft.Maui.Cli.UnitTests/Microsoft.Maui.Cli.UnitTests.csproj --filter "FullyQualifiedName~ProfileCommandTests" --logger "console;verbosity=minimal"`
- `./eng/common/dotnet.sh build src/Cli/Cli.slnf --no-restore --configuration Release`

## Issues

Fixes #385.

Addresses only the supported iOS portion of #109. Mac Catalyst remains separate future work.

PR #492, opened 2026-09-14
Fix Windows startup profiler dnx fallback

### LINKED ISSUE #383: Android profiling cannot start dnx fallback on Windows
## Description

On Windows, `maui profile startup` can fail after deploying and launching an Android app when the MAUI CLI falls back to invoking `dnx` for diagnostics tooling:

```text
Deploying and launching the app
Error: An error occurred trying to start process 'dnx' with working directory 'G:\SimpleHybridApp\SimpleApp'. The system cannot find the file specified.
```

The .NET SDK installs `dnx` on Windows as `C:\Program Files\dotnet\dnx.cmd`, but the profiler appears to start the extensionless `dnx` using `ProcessStartInfo` with `UseShellExecute = false`. That cannot resolve the `.cmd` wrapper.

## Reproduction

1. On Windows, use an SDK that supplies `C:\Program Files\dotnet\dnx.cmd` without a `dnx.exe`.
2. Ensure `dotnet-trace` and `dotnet-dsrouter` are not installed as global tools, so the CLI uses its `dnx` fallback.
3. Run an Android startup profile, for example:

```powershell
maui profile startup --project .\SimpleApp.csproj --framework net10.0-android --configuration Release --duration 00:00:15 --format speedscope --output .\startup.speedscope.json
```

4. After the app deploys and launches, the profile command fails trying to start `dnx`.

A direct .NET reproduction from the project directory also fails:

```csharp
new ProcessStartInfo("dnx", "--help")
{
    WorkingDirectory = @"G:\SimpleHybridApp\SimpleApp",
    UseShellExecute = false
}
```

Starting `dnx.cmd` explicitly succeeds.

## Expected behavior

The CLI should complete profiling with its fallback tooling on Windows, or report a prerequisite before deployment. It should resolve and launch `dnx.cmd` on Windows, or invoke an explicit supported diagnostics-tool path.

## Workaround

Installing the global diagnostics tools prevents the fallback:

```powershell
dotnet tool install --global dotnet-trace
dotnet tool install --global dotnet-dsrouter
```

## Environment

- Microsoft.Maui.Cli: `0.1.0-preview.12.26358.3+370e95b72f9b0656c3aa9a10d314847dd791319c`
- .NET SDK: `10.0.400-preview.0.26322.102`
- OS: Windows 11 (`10.0.26200`)
- Target framework: `net10.0-android`

Originally filed as dotnet/maui#36596; moving it here because the MAUI developer tools documentation identifies this repository as the experimental tooling issue tracker.

simonrozsival:
## Summary

Fix the startup profiler's Windows fallback when `dnx` resolves to the SDK's `.cmd` wrapper. The GitHub diff for this PR contains only the Windows `dnx` fallback fix for issue #383.

## Root cause

The diagnostics fallback validated `dnx` through `ProcessRunner.GetCommandPath("dnx")`, but later configured the child process with the unresolved command name `dnx`. On Windows, the resolved command can be a `.cmd` wrapper, which cannot be launched directly with `UseShellExecute = false` even though redirected standard input, output, and error are required by the profiler.

## Changes

- Resolve the actual `dnx` command path and use it for fallback process configuration.
- Preserve diagnostics-tool precedence: installed global tool, cached tool DLL, then `dnx`.
- Launch Windows `.cmd` wrappers through the Windows command processor with safely quoted arguments, while leaving redirected standard streams intact.
- Launch non-wrapper `dnx` paths directly through `ProcessStartInfo.ArgumentList`.
- Add focused tests for direct resolved paths and Windows `.cmd` wrapper configuration, including paths containing spaces.

## Stack relationship

This is layer 2 of a three-layer PR stack and depends on #491. Its base branch is `simonrozsival-profiling-injection-ios`; its GitHub diff contains only the Windows `dnx` fallback fix.

## Testing

- `./eng/common/dotnet.sh test src/Cli/Microsoft.Maui.Cli.UnitTests/Microsoft.Maui.Cli.UnitTests.csproj --filter 'FullyQualifiedName~Microsoft.Maui.Cli.UnitTests.ProfileCommandTests' --nologo`
- `./eng/common/dotnet.sh build src/Cli/Cli.slnf --no-restore --nologo`

## Issues

Fixes #383


src/Cli/Microsoft.Maui.Cli/Commands/ProfileCommandDiagnostics.cs
```diff
@@ -58,16 +58,52 @@ internal static void ConfigureDotnetToolStartInfo(ProcessStartInfo startInfo, st
...
+		ProcessStartInfo startInfo,
+		string dnxPath,
+		string packageId,
+		IReadOnlyList<string> toolArgs,
+		out string commandLine,
+		bool? isWindows = null)
+	{
+		var dnxArgs = new List<string> { "-y", packageId, "--" };
+		dnxArgs.AddRange(toolArgs);
+		if ((isWindows ?? OperatingSystem.IsWindows())
```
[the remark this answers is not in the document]
simonrozsival: Handled in `30713865`: Windows `.bat` wrappers now follow the same command-processor path as `.cmd` wrappers, and the focused test covers both extensions.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Cli/Microsoft.Maui.Cli/Commands/ProfileCommandDiagnostics.cs
```diff
@@ -58,16 +58,52 @@ internal static void ConfigureDotnetToolStartInfo(ProcessStartInfo startInfo, st
...
+			&& string.Equals(Path.GetExtension(dnxPath), ".cmd", StringComparison.OrdinalIgnoreCase))
+		{
+			startInfo.FileName = Environment.GetEnvironmentVariable("ComSpec") ?? "cmd.exe";
+			var commandProcessorLine = string.Join(" ", new[] { dnxPath }.Concat(dnxArgs).Select(QuoteForCommandProcessor));
```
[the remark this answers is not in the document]
simonrozsival: Handled in `30713865`: user-controlled paths and arguments are no longer serialized into the CMD command text. They are stored losslessly in dedicated environment variables and referenced through delayed expansion after CMD has parsed quotes and metacharacters; the focused test covers `%`, embedded quotes, carets, and shell metacharacters.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Cli/Microsoft.Maui.Cli/Commands/ProfileCommandDiagnostics.cs
```diff
@@ -58,14 +58,68 @@ internal static void ConfigureDotnetToolStartInfo(ProcessStartInfo startInfo, st
...
+			{
+				commandVariables[i] = $"__MAUI_CLI_DNX_VALUE_{i}";
+				startInfo.EnvironmentVariables[commandVariables[i]] = commandValues[i];
+			}
+
+			// Delayed expansion occurs after cmd.exe parses quotes and metacharacters.
+			var commandProcessorLine = string.Join(" ", commandVariables.Select(name => $"\"!{name}!\""));
+			startInfo.ArgumentList.Add("/d");
+			startInfo.ArgumentList.Add("/s");
+			startInfo.ArgumentList.Add("/v:on");
```
jfversluis, line 102: `/v:on` is inherited by `dnx.cmd`, which later forwards the arguments through `%*`. Valid values containing `!` can therefore be interpreted as delayed-expansion expressions and corrupted; the wrapper's `%~dp0` handling can be affected as well when the SDK path contains `!`. The current test covers several cmd metacharacters but not `!`, and only inspects `ProcessStartInfo`. Could we avoid executing the `.cmd` wrapper and invoke its underlying `dotnet.exe dnx` command directly, then add an execution-level Windows test with `!` in both a path and an argument?

jfversluis (commented):
I found one Windows command-wrapper correctness issue in this layer.

src/Cli/Microsoft.Maui.Cli/Commands/ProfileCommandDiagnostics.cs
simonrozsival: Fixed in `741c0ac4` by bypassing the wrapper: resolved `.cmd`/`.bat` paths now launch the sibling `dotnet.exe` directly with `dnx` and all values in `ArgumentList`, so no CMD expansion is involved. I also added a Windows execution-level round-trip test that runs from an SDK-style path containing `!` and passes an output argument containing `!`.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Cli/Microsoft.Maui.Cli.UnitTests/ProfileCommandTests.cs
```diff
@@ -1029,6 +1030,145 @@ public void CanUseDiagnosticsTooling_MissingRequiredToolWithoutDnx_ReturnsFalse(
...
+			Assert.True(process.ExitCode == 0, standardError);
+			var actualArgs = System.Text.Json.JsonSerializer.Deserialize<string[]>(standardOutput);
+			Assert.NotNull(actualArgs);
+			Assert.Equal(
+				["dnx", "-y", "dotnet-trace", "--", "collect", "--output", outputPath],
+				actualArgs);
+		}
+		finally
+		{
+			Directory.Delete(tempDirectory, recursive: true);
```
jfversluis, line 1175: Windows CI fails here with `UnauthorizedAccessException: Access to the path 'dotnet.exe' is denied`. The `using var process` declaration keeps the launched helper process alive until the method scope exits, but this `finally` tries to recursively delete its executable directory first. Please dispose the process before entering cleanup (for example, use a scoped `using (...) { ... }` block around execution), then delete the directory and rerun CI. The underlying `dotnet.exe dnx` fix itself looks correct.

jfversluis (dismissed):
The product-code fix addresses the original review finding, but the new Windows regression test currently fails CI because its cleanup runs before the launched process is disposed.

src/Cli/Microsoft.Maui.Cli.UnitTests/ProfileCommandTests.cs
simonrozsival: Fixed in `3e3adfb1`: the launched helper process now has an explicit scoped `using` block, so it is disposed before the `finally` recursively deletes the apphost directory. The Windows execution-level `!` path and argument coverage remains unchanged.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Cli/Microsoft.Maui.Cli.UnitTests/ProfileCommandTests.cs
```diff
@@ -1029,6 +1030,147 @@ public void CanUseDiagnosticsTooling_MissingRequiredToolWithoutDnx_ReturnsFalse(
...
+				""");
+			File.WriteAllText(
+				Path.Combine(tempDirectory, "Program.cs"),
+				"""
+				using System;
+
+				Console.Write(System.Text.Json.JsonSerializer.Serialize(args));
+				""");
+
+			var dotnetPath = ProcessRunner.GetCommandPath("dotnet");
```
jfversluis: This test resolves `dotnet` from the process-wide `PATH`, while `AndroidProviderTests` temporarily sets `PATH` to an empty string and xUnit can run the classes in parallel. On #493's full Windows run this returned null here, even though this PR happened to pass on its prior run. Please make the fixture independent of mutable global `PATH` (or place the environment-mutating tests in a non-parallel collection) so the regression test is deterministic across the stacked branches.

jfversluis (changes requested):
The cleanup fix is correct and #492 is green, but the same regression test failed after the stacked branch ran the full suite, exposing a parallel-test PATH race.

src/Cli/Microsoft.Maui.Cli.UnitTests/ProfileCommandTests.cs
simonrozsival: Fixed in `48d201d7`: the fixture now derives the active `dotnet.exe` from `RuntimeEnvironment.GetRuntimeDirectory()` instead of resolving through process-wide `PATH`, so concurrent environment mutation cannot affect it. The Windows execution-level regression coverage remains unchanged.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

jfversluis (approved):
Re-reviewed the follow-up fixes. The dnx wrapper path now bypasses CMD safely, process cleanup is correctly scoped, and the execution fixture derives the active dotnet host independently of mutable PATH. macOS and Windows CI are green.

PR #493, opened 2026-09-14
Fix physical Android startup trace lifecycle

### LINKED ISSUE #324: CLI: startup profiling does not work with physical devices
I ran into a problem where startup profiling on Android works just fine with an emulator, but it does not successfully connect to a physical device:

```
> maui profile startup --project MauiApp.csproj -f net11.0-android --format mibc --stopping-event-provider-name Microsoft.Mau
i.StartupProfiling --stopping-event-event-name StartupComplete
Info: Executing: /usr/bin/xcrun simctl list devices --json --json-output=/var/folders/gz/0_5rr5zd5tgf68xmn8wrgv2m0000gn/T/tmpx2lgyq.tmp
Info: Found 58 simulator device(s).
ℹ Project: src/maui/MauiApp.csproj
ℹ Framework: net11.0-android
ℹ Configuration: Release
ℹ Device: SM_G991U1 (R3CR20CD53A)
ℹ Format: mibc
ℹ Output: src/maui/MauiApp_20260611_172419.mibc
ℹ Raw trace companion: src/maui/MauiApp_20260611_172419.nettrace
ℹ Diagnostic port: 9000
ℹ Building the app (1:52.5s)
ℹ Deploying and launching the app (1:47.4s)
ℹ Waiting for the configured stopping event. Press Enter to stop early.
Error [E1001]: dotnet-trace exited with code 2.
  For finer control over the dotnet-dsrouter options, run it separately and
connect to it using -p
 
WARNING: dotnet-dsrouter is a development tool not intended for production
environments.
 
How to connect current dotnet-dsrouter pid=92590 with android device and
diagnostics tooling.
Build and run your application on android device such as:
[Default Tracing]
dotnet build -t:Run -c Release -p:DiagnosticAddress=127.0.0.1
-p:DiagnosticPort=9000 -p:DiagnosticSuspend=false
-p:DiagnosticListenMode=connect
[Startup Tracing]
dotnet build -t:Run -c Release -p:DiagnosticAddress=127.0.0.1
-p:DiagnosticPort=9000 -p:DiagnosticSuspend=true -p:DiagnosticListenMode=connect
See https://learn.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-dsrouter
for additional details and examples.
 
info: dotnet-dsrouter-92590[0]
      Starting dotnet-dsrouter using pid=92590
info: dotnet-dsrouter-92590[0]
      Shutting down due to error: Address already in use
info: dotnet-dsrouter-92590[0]
      Stopping IPC server
(/var/folders/gz/0_5rr5zd5tgf68xmn8wrgv2m0000gn/T/dotnet-diagnostic-dsrouter-925
90-1781198879-socket) <--> TCP server (127.0.0.1:9001) router.
info: dotnet-dsrouter-92590[0]
      Router stopped.
[ERROR] Failed to launch dsrouter: android. Make sure that dotnet-dsrouter is
not already running. You can connect to an already running dsrouter with -p.
```

I wonder if we're misconfiguring dotnet-trace or dotnet-dsrouter or if there is simply a problem with adb reverse?

### LINKED ISSUE #386: maui profile startup: dsrouter fails to bind 127.0.0.1:9001 (EADDRINUSE) even when ports 9000/9001 are free
### Description

`maui profile startup --format mibc` fails at the tracing step with `dotnet-trace exited with code 2`, because the `dotnet-dsrouter` it launches cannot bind its TCP server on `127.0.0.1:9001` — **even when ports 9000 and 9001 are confirmed free before the run**.

The build, deploy, and launch all succeed. The failure happens only once tracing starts ("Trace is running. Press Enter to stop…"), and dsrouter immediately shuts down with an `EADDRINUSE` (socket-in-use) error on `127.0.0.1:9001`, then the whole command aborts.

Because the ports are verified empty beforehand (`Get-NetTCPConnection -LocalPort 9000,9001` returns nothing, and no `dotnet-dsrouter` / `dotnet-trace` process is running), this is **not** a leftover/stale-process collision. The command appears to collide with a dsrouter it started itself — i.e. dsrouter (or dotnet-trace's implicit dsrouter) is being started more than once on the same fixed port `9001`, with no way to configure that port.

### Steps to Reproduce

1. Ensure ports 9000 and 9001 are free and no `dotnet-dsrouter` / `dotnet-trace` process is running:
   ```powershell
   Get-NetTCPConnection -LocalPort 9000,9001 -ErrorAction SilentlyContinue   # returns nothing
   Get-Process dotnet-dsrouter,dotnet-trace -ErrorAction SilentlyContinue     # returns nothing
   ```
2. Install the diagnostics global tools (required to work around the dnx-fallback issue):
   ```
   dotnet tool install --global dotnet-trace
   dotnet tool install --global dotnet-dsrouter
   ```
3. Run against a `net11.0-android` (CoreCLR) app in Release:
   ```
   maui profile startup --format mibc
   ```

### Expected Behavior

The command starts dsrouter once, binds `127.0.0.1:9001`, captures the startup trace, and produces the `.mibc` / `.nettrace` output.

### Actual Behavior

Build, deploy and launch succeed. Tracing then fails:

```
? Building the app (8:59.7s)
? Deploying and launching the app (7:45.5s)
? Trace is running. Press Enter to stop and finalize the trace output.
Error [E1001]: dotnet-trace exited with code 2.
  For finer control over the dotnet-dsrouter options, run it separately and connect to it using -p

WARNING: dotnet-dsrouter is a development tool not intended for production environments.

How to connect current dotnet-dsrouter pid=45540 with android device and diagnostics tooling.
...
info: dotnet-dsrouter-45540[0]
      Starting dotnet-dsrouter using pid=45540
info: dotnet-dsrouter-45540[0]
      Shutting down due to error: <socket address already in use>
info: dotnet-dsrouter-45540[0]
      Stopping IPC server (dotnet-diagnostic-dsrouter-45540) <--> TCP server (127.0.0.1:9001) router.
info: dotnet-dsrouter-45540[0]
      Router stopped.
[ERROR] Failed to launch dsrouter: android. Make sure that dotnet-dsrouter is not already running. You can connect to an already running dsrouter with -p.
```

(The socket message is localized; in English it is "Only one usage of each socket address … is normally permitted.")

### Notes / Analysis

- Ports `9000` and `9001` were verified free immediately before the run, so this is not a stale/orphaned-process collision.
- The dsrouter TCP port (`9001`) appears to be hard-coded — there is no CLI flag to change it, and the diagnostic port flag only controls `9000`.
- The error text ("Make sure that dotnet-dsrouter is not already running. You can connect to an already running dsrouter with `-p`") and the hint "run it separately and connect to it using `-p`" suggest the CLI may be starting dsrouter twice (or dotnet-trace's implicit dsrouter races with the one the CLI launches), causing the process to collide with itself on `9001`.
- Possibly related environment to #385 (same app / same `net11.0-android` CoreCLR setup).

### Version Information

- `microsoft.maui.cli` **0.1.0-preview.12.26358.3**
- `dotnet-trace` / `dotnet-dsrouter` global tools **9.0.661903**
- Target framework: `net11.0-android` (CoreCLR runtime), Release configuration
- Device: physical Pixel 9 Pro
- Host OS: Windows 11

### LINKED ISSUE #422: maui profile startup: dsrouter TCP 9001 self-collision with default --diagnostic-port 9000 (Android/USB)
### Description

`maui profile startup` on a physical Android device fails during trace collection with a TCP port collision on **9001**:

```
Error [E1001]: dotnet-trace exited with code 2.
info: dotnet-dsrouter-...[0]
      Shutting down due to error: Only one usage of each socket address (protocol/network address/port) is normally permitted.
info: dotnet-dsrouter-...[0]
      Stopping IPC server (...) <--> TCP server (127.0.0.1:9001) router.
[ERROR] Failed to launch dsrouter: android. Make sure that dotnet-dsrouter is not already running.
```

With the **default `--diagnostic-port 9000`**, the CLI sets up an `adb reverse` mapping onto host port **9001**, which is exactly where the `dotnet-trace --dsrouter android` dsrouter binds its TCP server → they collide and the trace never starts. Nothing is holding 9001 between runs; it is a self-collision within a single invocation.

### Workaround

Pass **`--diagnostic-port 9100`** (any value whose derived +1 port ≠ 9001). With that, the dsrouter binds 9001 cleanly and collection proceeds.

### Steps to reproduce

1. Physical Android device over USB.
2. `maui profile startup --project <app>.csproj -f net11.0-android -d <device> --format mibc -c Release --duration 00:00:20`
3. Observe the dsrouter socket-in-use error on 9001.

### Expected

The CLI should pick non-colliding ports (or offset its `adb reverse` off the dsrouter's own TCP port) so the default invocation works.

### Environment

- Host: Windows 11
- .NET SDK: `11.0.100-preview.6.26359.118`
- maui CLI (`microsoft.maui.cli`): `0.1.0-preview.12.26358.3`
- `dotnet-trace` / `dotnet-dsrouter`: `9.0.661903`
- App: .NET MAUI–embedding Avalonia app, TFM `net11.0-android`, CoreCLR runtime
- Device: Google Pixel 9 Pro (physical, USB)

### LINKED ISSUE #424: maui profile startup: hardcoded 15s finalize timeout (s_traceStopTimeout) not configurable -> --format mibc never completes
### Description

After solving the port collision (see companion issue) with `--diagnostic-port 9100`, `maui profile startup` connects and collects the trace, but **always fails at finalize** with:

```
? Trace is running. It will stop automatically after 00:00:00:20 unless you press Enter sooner.
? Stopping trace and finalizing output...
Error [E1001]: dotnet-trace did not exit within 15s after the stop request.
```

No `.nettrace`/`.mibc` is produced. The 15 s wait appears to be the hardcoded `ProfileCommand.s_traceStopTimeout = TimeSpan.FromSeconds(15)`, with **no CLI option or environment variable to override it**. For a real app, dotnet-trace's stop/finalize (rundown + stream flush over dsrouter/USB) legitimately takes longer than 15 s, so `--format mibc` can never complete.

Notably, the failure is **independent of `--duration`** — it fails even at `--duration 00:00:08`, so it is the finalize/flush, not the amount of collected data.

### Request

- Make the stop/finalize timeout configurable (option or env var), and/or
- Increase / auto-scale the default, and/or
- Don't hard-fail: if the child is still finalizing, keep waiting (or surface the partially written trace).

### Steps to reproduce

1. Physical Android device over USB, non-trivial app.
2. `maui profile startup --project <app>.csproj -f net11.0-android -d <device> --format mibc -c Release --duration 00:00:20 --diagnostic-port 9100`
3. Collection runs, then finalize aborts at 15 s.

### Environment

- Host: Windows 11
- .NET SDK: `11.0.100-preview.6.26359.118`
- maui CLI (`microsoft.maui.cli`): `0.1.0-preview.12.26358.3`
- `dotnet-trace` / `dotnet-dsrouter`: `9.0.661903`
- App: .NET MAUI–embedding Avalonia app, TFM `net11.0-android`, CoreCLR runtime
- Device: Google Pixel 9 Pro (physical, USB)

simonrozsival:
## Summary

Fix physical Android startup and manual profiling sessions that collect through `dotnet-trace`/`dotnet-dsrouter`. The CLI now owns the router lifecycle explicitly, keeps the diagnostics and exit-control channels independent, and gives trace rundown/finalization a configurable timeout.

## Root causes

Physical Android collection previously used `dotnet-trace --dsrouter android`. That implicit router hardcodes host TCP port 9001. With the default diagnostic port 9000, the CLI also allocated 9001 for its profiling-helper exit-control channel, so the router and CLI could collide with themselves. The implicit router also hid the router process, IPC endpoint, adb mapping, and shutdown lifecycle from the CLI, making cleanup and arbitrary `--diagnostic-port` values unreliable.

Trace stopping also used a fixed 15-second wait after the stop sequence. Large traces and slower physical devices could still be flushing/running rundown, while callers could not tune the finalization window or retain collector output in a deterministic timeout failure.

## Port architecture

For physical Android sessions using `dotnet-trace`:

- **N**: device diagnostic port requested by `--diagnostic-port`
- **N+1**: unique host TCP port owned by explicit `dotnet-dsrouter server-server`
- **N+2**: independent CLI exit-control channel

The CLI launches dsrouter with a unique IPC endpoint, `--tcp-server 127.0.0.1:N+1`, and `--forward-port Android`. Dsrouter owns the diagnostic adb reverse from device port N to host port N+1. `dotnet-trace` connects to the explicit IPC endpoint with `--diagnostic-port <endpoint>,connect`, so it does not launch a second implicit router. The CLI retains its host reservation for N until cleanup so another concurrent session cannot select the same device diagnostic port and replace the active adb mapping.

## Changes

- Reserve physical-device diagnostics, dsrouter TCP, and exit-control ports as an atomic N/N+1/N+2 set.
- Clear stale diagnostic mappings before dsrouter startup and independently configure the exit-control mapping.
- Start and monitor an explicit dsrouter process with a unique IPC endpoint.
- Clean trace/router processes, reservations, adb mappings, exit-control listeners, and Unix IPC socket paths on success, setup failure, launch failure, cancellation, and finalization failure.
- Add `--trace-stop-timeout` to `profile startup` and `profile manual`, defaulting to two minutes and rejecting non-positive values.
- Apply the timeout to the complete stop, interrupt, rundown, flush, and finalization window while preserving collector output in hard failures, including collector-managed duration and stopping-event finalization.
- Document manual profiling and the independent finalization timeout.
- Add regression coverage for option/default/validation behavior, explicit dsrouter arguments and IPC endpoints, physical-device port allocation, explicit dotnet-trace endpoint arguments, timeout failures with collector output, collectors that exit before the interrupt fallback, and timed collector finalization.

## Compatibility and scope

Android emulator and iOS collection continue using their existing implicit dsrouter modes. Runtime-owned Android `.nettrace` collection does not launch explicit dsrouter and keeps its prior adjacent exit-control allocation. This layer contains only the remaining Android physical-device lifecycle/finalization work; it does not duplicate the profiling injection changes or Windows dnx fallback from the lower layers.

## Stack relationship

This is layer 3 (top) of the profiling fix stack. It targets and depends on #492, which transitively depends on #491.

## Testing

- `ProfileCommandTests`: 118 passed
- Full `src/Cli/Cli.slnf` build: succeeded with 0 warnings and 0 errors
- Full CLI unit suite before the review follow-up: 889 passed, 2 failed, 0 skipped
  - `DevFlowCliCommandTests.UiGesture_ReportsWhichTierHandledIt`
  - `DevFlowCliCommandTests.UiGesture_Pinch_SendsScaleToGestureRoute`
- Both DevFlow gesture failures reproduced unchanged on the original layer-2 base SHA `43fa3ae7a21c56bc2e7c87f6f82b050e4a8a2602`.

## Issues

Closes #324
Closes #386
Closes #422
Closes #424


src/Cli/Microsoft.Maui.Cli/Commands/ProfileTraceLifecycle.cs
```diff
@@ -103,6 +112,61 @@ internal static async Task<bool> WaitForCompletionAsync(
...
+		Task processWaitTask,
+		TimeSpan traceStopTimeout,
+		IOutputFormatter formatter,
+		bool useJson,
+		bool verbose)
+	{
+		using var timeoutSource = new CancellationTokenSource(traceStopTimeout);
+		try
+		{
+			await RequestStopAsync(traceProcess.Process, formatter, useJson, verbose, timeoutSource.Token);
```
jfversluis, line 126: This configurable timeout still does not give an acknowledged rundown the requested window: `RequestStopAsync` sends SIGINT after the fixed five-second interrupt delay whenever the process remains alive, even if `dotnet-trace` has already emitted the finalization signal. Exit code 130 is subsequently accepted as success, so a long rundown can be interrupted and an incomplete trace reported as successful. Once finalization has been acknowledged, can we wait for the full `traceStopTimeout` and reserve SIGINT for a stop request that was never acknowledged? A test where finalization starts before five seconds but completes afterward would cover this.

src/Cli/Microsoft.Maui.Cli/Commands/ProfileCommandPortRouter.cs
```diff
@@ -75,29 +75,54 @@ internal static async Task<ReservedProfilePorts> ReserveProfilePortsAndConfigure
...
 				if (transport.RequiresManualExitControlPortRouting)
 				{
+					if (transport.RequiresExplicitDsrouter)
+					{
+						ProfileCommandProcessHelpers.WriteVerbose(
+							formatter,
+							useJson,
+							verbose,
+							$"Clearing any stale adb mapping for the device diagnostic port {port} before dotnet-dsrouter takes ownership.");
+						await RemoveAdbPortRoutingAsync(device, formatter, useJson, verbose, port);
```
jfversluis, line 123: This removes any existing device mapping for port N without proving that it is stale or owned by this profiling session. A different active debugger/profiler can legitimately own `tcp:N`, and cleanup later removes the port again without checking whether ownership changed. Could we query `adb reverse --list`, treat a pre-existing mapping as a collision and advance to the next port set, then remove the diagnostic mapping only when it still matches the exact N -> N+1 mapping created by this session?

jfversluis (commented):
I found two lifecycle/ownership issues in this layer.

src/Cli/Microsoft.Maui.Cli/Commands/ProfileTraceLifecycle.cs
simonrozsival: Fixed in `8faae045`. Stop handling now distinguishes acknowledged finalization from the SIGINT fallback, gives acknowledged rundown a fresh full `traceStopTimeout`, and accepts exit 130 only when the root collector was actually signaled. The new regression test starts finalization before the fallback delay and verifies clean completion afterward without interruption.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Cli/Microsoft.Maui.Cli/Commands/ProfileCommandPortRouter.cs
simonrozsival: Fixed in `8faae045`. Port selection now parses `adb reverse --list` and treats any existing device mapping as a collision instead of deleting it; exit mapping creation uses `--no-rebind`, and cleanup is state-tracked and removes only a single exact expected mapping. The diagnostic path also rechecks immediately before dsrouter starts while retaining the required dsrouter-owned `--forward-port Android` behavior.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Cli/Microsoft.Maui.Cli.UnitTests/ProfileCommandTests.cs
```diff
@@ -51,11 +52,285 @@ public void ProfileStartupCommand_HasExpectedOptions()
...
+			await process.WaitForExitAsync();
+		}
+	}
+
+	[Fact]
+	public async Task StopAndWaitForFinalizationAsync_ReturnsWhenCollectorExitsBeforeInterruptDelay()
+	{
+		var startInfo = new ProcessStartInfo
+		{
+			FileName = OperatingSystem.IsWindows() ? "powershell.exe" : "/bin/sh",
```
jfversluis: Both new lifecycle tests use `powershell.exe` as the fake collector but give the entire process only 1-2 seconds to exit. On Windows CI, PowerShell startup itself took roughly 3 seconds, so the 100/200 ms sleep processes missed those deadlines and both tests failed. Please use a lightweight deterministic helper process (ideally a small existing test fixture/apphost) or make the synchronization event-driven rather than relying on sub-second wall-clock timing; simply increasing the timeout would leave these tests flaky under runner load.

jfversluis (changes requested):
The product fixes still look correct, but the new Windows lifecycle tests are timing out because their timeout assumptions exclude PowerShell startup time.

src/Cli/Microsoft.Maui.Cli.UnitTests/ProfileCommandTests.cs
simonrozsival: Fixed in `7de3de7f`. The lifecycle tests now use a small cross-platform .NET helper executable copied through a project reference, with explicit `ready` and `finalizing` stdout signals. This removes PowerShell and shell startup timing from the assertions instead of increasing the timeouts.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

simonrozsival:
@jfversluis All of your review feedback has been addressed, including replacing the PowerShell timing tests with an event-driven .NET file app. The review threads are resolved and macOS/Windows CI is green. I dismissed the stale changes request and re-requested your review—could you please take another look and merge if it looks good?

jfversluis (approved):
Re-reviewed the follow-up commits. The lifecycle tests now use explicit ready/finalizing signals rather than shell startup timing, and interrupted descendant collectors are tracked correctly when the wrapper exits first. The earlier timeout and adb ownership findings remain addressed, and macOS/Windows CI are green.

PR #498, opened 2026-09-16
fix: don't pack the Blazor project's PRI (MSB3030)

fourwhey:
Microsoft.Maui.DevFlow.Blazor cannot be referenced by any consuming project. Adding the package fails the build with MSB3030, looking for lib/<tfm>/Microsoft.Maui.DevFlow.Blazor/wwwroot/chobitsu.js.

MakePri indexes wwwroot/chobitsu.js into this project's PRI, while the Razor SDK packs that file into staticwebassets/ and the PRI into lib/. In the consumer, WindowsAppSDK's AddPriPayloadFilesToCopyToOutputDirectoryItems expands every referenced .pri and resolves its payload against the .pri's own folder, so it demands a path the package does not contain.

The PRI's entire contents is that single wwwroot entry, and chobitsu.js already reaches consumers through the static web asset pipeline, so the project's PRI has no legitimate content. Stop packing it.

_ExcludeWwwrootFromPri, added for #66, does not work: removing the Content items before GetMrtPackagingOutputs does not change what MakePri indexes, and the PRIs shipped in 0.1.0-preview.12.26368.2 and 0.1.0-preview.12.26421.1 are identical. AppxGeneratePriEnabled=false is not an alternative either, as pack still expects the file and fails NU5026.


fourwhey:
@dotnet-policy-service agree

